PR #2, opened 2023-05-03
MySQL CDC adapter and Kafka upserts

ryannedolan:
New MySQL CDC adapter via MySqlCdcSchemaFactory. Leverages flink-sql-connector-mysql-cdc, which runs Debezium under-the-hood.

RawKafkaSchemaFactory now uses upsert-kafka connector and CSV format in order to support MySQL -> Kafka pipelines. Without a schema registry, the schema is still hard-coded as PAYLOAD, KEY.

Added a MySQL database to the dev environment, along with sample subscriptions.

## Testing ##

Not all tables in the example MySQL database work due to unsupported data types. But `products_on_hand` works:

```
$ ./bin/hoptimator
> SELECT * FROM INVENTORY."products_on_hand" LIMIT 5;
+------------+----------+
| product_id | quantity |
+------------+----------+
| 106        | 0        |
| 105        | 5        |
| 108        | 2        |
| 107        | 44       |
| 102        | 8        |
+------------+----------+
5 rows selected (2.159 seconds)

```
The `LIMIT` clause is required, as this is an unbounded CDC stream.

We can deploy a subscription that writes change events to Kafka:

```
$ kubectl apply -f products.yaml
subscription.hoptimator.linkedin.com/companies created

$ kubectl get subscriptions
NAME        DB         SQL
products    RAWKAFKA   SELECT "quantity", "product_id" AS KEY FROM INVENTORY."products_on_hand"
```

...and then query the resulting Kafka topic:

```
> SELECT * FROM RAWKAFKA."products" LIMIT 5;
+---------+-----+
| PAYLOAD | KEY |
+---------+-----+
| 5       | 109 |
| 44      | 107 |
| 2       | 108 |
| 18      | 103 |
| 8       | 102 |
+---------+-----+
5 rows selected (3.859 seconds)

```


deploy/hoptimator-pod.yaml
```diff
@@ -8,3 +8,11 @@ spec:
...
     imagePullPolicy: Never
     command: ["sleep", "infinity"]
+    volumeMounts:
+    - name: config-volume
+      mountPath: /etc/config
+  volumes:
+  - name: config-volume
+    configMap:
+      name: hoptimator-configmap
+
```
ehoner, line 18: Are both a `deployment` and `pod` spec necessary?

Makefile
```diff
@@ -1,32 +1,40 @@
...
 
-deploy:
+deploy: deploy-config
 	kubectl apply -f ./deploy/
 
+deploy-samples:
+	kubectl apply -f ./deploy/samples
+
+deploy-config:
+	kubectl create configmap hoptimator-configmap --from-file=model.yaml=test-model.yaml --dry-run=client -o yaml | kubectl apply -f -
```
ehoner, line 35: nit: Should the default file be called `test-model.yaml`? Is it setup like this to keep testing/local development simple?

It was removed from `Dockerfile` (this change), but it is still referenced as `test-model.yaml` in [HoptimatorOperatorApp](https://github.com/linkedin/Hoptimator/blob/041321a0c670ceb793e143263c45d4d461959d13/hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/HoptimatorOperatorApp.java#L31).

ehoner (changes requested):
Minor comments, I am not strongly committed to them.

deploy/hoptimator-pod.yaml
ryannedolan: The deployment is for the operator, while the pod is basically a jump box. They are the same image, and it would be reasonable to combine them into one deployment (i.e. we could just jump to one of the operators). Not sure which is more ergonomic yet.

ehoner (approved):

PR #4, opened 2023-05-08
Integration tests and workflow

ryannedolan:
Tweaks to enable running integration tests in the form of a SQL script. A similar SQL script is used for a new readiness probe, which enables the workflow to wait until the cluster is ready.

## Testing

```
$ make integration-tests
...
PASS
```

Also, the GitHub workflow below is passing.


ehoner (approved):
This is great. The consistency between the local environment and integration-testing is spectacular. The only difference I see is with java. Is this something that gradle will enforce/manage?

PR #9, opened 2023-05-11
Add simple MySQL CDC integration test

ryannedolan:
The sample subscriptions include a MySQL CDC -> Kafka pipeline. For a simple integration test, we can try querying the table and the resulting Kafka topic.

We can do something similar as a readiness probe, which lets us wait to start the integration tests until after Kafka etc is ready. Eventually, the integration tests should check for correctness etc, while the readiness probes can remain trivial.

## Testing

Integration tests show the CDC pipeline working:

```
--->%---
9/10         -- MySQL CDC -> Kafka
10/10        SELECT * FROM RAWKAFKA."products" LIMIT 1;
--->%---
+---------+-----+
| PAYLOAD | KEY |
+---------+-----+
| 3       | 101 |
+---------+-----+
```


ryannedolan:
I tripped over a couple bugs in the integration test workflow. ftr:

- The K8s cluster was running out of CPU quota. I reduced Flink's CPU requests to 0.1, which seems to be the lowest I can go.
- The readiness probe cannot include subscription tests -- subscriptions aren't deployed until _after_ readiness.

bin/hoptimator
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-kubectl exec -it hoptimator -c hoptimator -- ./hoptimator -n "" -p "" -u "jdbc:calcite:model=/etc/config/model.yaml" "$@"
+kubectl exec -it hoptimator -c hoptimator -- ./hoptimator --isolation=TRANSACTION_NONE -n "" -p "" -u "jdbc:calcite:model=/etc/config/model.yaml" -nn hoptimator "$@"
```
ehoner, line 3: Is the `-n` flag used twice? The second occurrence is `-nn`. fwiw, I couldn't figure out how/if it affected the tests.

ehoner (approved):
Just the one question, if it isn't an issue, go ahead and merge.

PR #14, opened 2023-05-23
Add insert-into command for testing purposes

ryannedolan:
A new CLI command `!insert into` enables writing to existing tables for testing purposes. The command creates an ephemeral job which runs in-process in the background and returns "SUCCESS" (barring any validation errors). This is useful for testing multi-hop pipelines without deploying them as "subscriptions".

A new integration test covers this use-case.


hoptimator-cli/src/main/java/com/linkedin/hoptimator/HoptimatorCliApp.java
```diff
@@ -264,6 +266,97 @@ public boolean echoToFile() {
...
+        sql = sql.substring(1);
+      }
+
+      if (sql.startsWith("insert into")) {
+        sql = sql.substring("insert into".length() + 1);
+      }
+
+      String connectionUrl = sqlline.getConnectionMetadata().getUrl();
+      try {
+        String[] parts = sql.split("[sS][eE][lL][eE][cC][tT]");
```
ehoner, line 314: nit: This could be case-insensitive regex. I don't have strong preferences for readability either way though.
```java
jshell> String stmt = "SelEcT * FROM RAWKAFKA.\"test-sink\" LIMIT 5;"
stmt ==> "SelEcT * FROM RAWKAFKA.\"test-sink\" LIMIT 5;"

jshell> stmt.split("(?i)select")
$7 ==> String[2] { "", " * FROM RAWKAFKA.\"test-sink\" LIMIT 5;" }
```

ehoner (approved):
Fix or merge.

hoptimator-cli/src/main/java/com/linkedin/hoptimator/HoptimatorCliApp.java
ryannedolan: Neat! Will fix.

Issue #17, opened 2023-05-24
Command-line arguments for hoptimator-operator


ryannedolan:
Currently `hoptimator-operator` accepts one argument, the path to a Calcite "model file". Remaining properties are hard-coded, including the K8s namespace ("default"). We need a way to provide a few additional command-line arguments, e.g. `--namespace my-namespace`, along with sensible defaults.

Additionally, individual controllers/reconcilers may need to be configured with their own properties. Controllers are pluggable, so we need a way to pass arbitrary key-value properties via command-line arguments, e.g. `--property key=value`. Controllers should be able to see these properties via the Operator object, e.g. `Operator.properties().get("key")`.

stevenewald:
Would you mind giving an example of how to test this?

ryannedolan:
Hmm, this turns out to be difficult to test, cuz operator properties really shouldn't be observable from the SQL layer. For example, say we want the kafka topic controller to have a configurable partition count or retention period for new topics. Those are specific to Kafka, and have no bearing on the results you'd see from a query. So we could pass those properties on the command line to the operator, and the kafka controller could change its behavior accordingly, but you wouldn't be able to test that from the SQL CLI.

Or e.g. if you change the k8s namespace used by the operator,  or the service account used by the operator, that sort of thing should be invisible to the SQL layer. That's by design!

So I don't think we can test this within SQL, unless we do something artificial. I'd rather we write a unit test (if there is anything unit-testable) or just test it manually. I'd probably add a log line to the operator that spits out all the properties, and then just look for my properties to show up. For a manual test, you can modify `hoptimator-operator-deployment.yaml` to include additional command-line arguments.

One thing that's worth manually testing is the k8s namespace. It's not uncommon to run an operator in one namespace but have it control resources in another namespace. The dev environment uses the "default" namespace, but it should be possible to modify `hoptimator-operator-deployment.yaml` to pass `--namespace foo` to the operator. That would enable you to create KafkaTopics, Subscriptions, etc in the foo namespace, e.g. `kubectl apply -n foo -f deploy/samples`, and the operator should see them. If we can document that this works in the PR, we don't need to bother automating a test for it.

PR #18, opened 2023-05-24
Add pluggable controllers via ControllerProvider

ryannedolan:
Previously, a new adapter would require writing (or finding) an entirely new operator, or using one of the built-in reconcilers (e.g. KafkaTopicReconciler). This made it difficult to extend Hoptimator with new adapters or to customize the way the operator works.

Instead, we can make the operator pluggable, just as we have made the planner and catalog pluggable. A single adapter can now bundle catalog, planner, and operator plugins in the same place. This means we can drop in a custom Kafka adapter, for example, and get completely different logic for creating/deleting topics. (This turns out to be necessary for internal use-cases.)

An interesting side-effect of this is that the CLI now has access to control plane logic, which could be used to create/delete tables without running an operator.


hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/ControllerService.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.stream.Collectors;
+
+/** Loads controller plugins via the ControllerProvider Service Provider Interface. */
+public final class ControllerService {
+
+  public static Collection<ControllerProvider> providers() {
+    ServiceLoader<ControllerProvider> loader = ServiceLoader.load(ControllerProvider.class);
+    List<ControllerProvider> providers = new ArrayList<>();
+    loader.iterator().forEachRemaining(x -> providers.add(x));
+    return providers;
```
ehoner, line 20: nit: 
```java
// loader variable isn't required
// change providers to lambda
List<ControllerProvider> providers = new ArrayList<>(); 
ServiceLoader.load(ControllerProvider.class).iterator().forEachRemaining(providers::add);
return providers;
```

FI alternative (requires JavaCompile > 8).
```java
return ServiceLoader.load(ControllerProvider.class)
    .stream()
    .map(ServiceLoader.Provider::get)
    .collect(Collectors.toList());
```

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/Operator.java
```diff
@@ -21,27 +21,36 @@
...
   private final Map<String, ApiInfo<?, ?>> apiInfo = new HashMap<>();
+  private final Properties properties;
 
-  public Operator(String namespace, ApiClient apiClient, SharedInformerFactory informerFactory) {
+  public Operator(String namespace, ApiClient apiClient, SharedInformerFactory informerFactory,
+      Properties properties) {
     this.namespace = namespace;
     this.apiClient = apiClient;
     this.informerFactory = informerFactory;
+    this.properties = properties;
```
ehoner, line 44: I think it is better to avoid potentially `null` fields. Add test for input or require non-null. ie. 
```
# test
this.properties = properties == null ? new Properties() : properties;

# validate
this.properties = Validate.notNull(properties);
```

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/HoptimatorOperatorApp.java
```diff
@@ -6,29 +6,42 @@
...
+    this.properties = properties;
+  }
+
   public static void main(String[] args) throws Exception {
     if (args.length != 1) {
       throw new IllegalArgumentException("Missing model file argument.");
     }
-    HoptimatorPlanner.Factory plannerFactory = HoptimatorPlanner.Factory.fromModelFile(args[0],
-        new Properties());
+    new HoptimatorOperatorApp(args[0], "default", new Properties()).run();
```
ehoner, line 39: Is it useful to allow overrides from the command line? If `new Properties()` is replaced with `System.getProperties()`, then `-Dhoptimator.key=value` settings would be honored with `run.sh`. Execution would be similar to:
```sh
$ JAVA_OPTS=-Dhoptimator.cli=value hoptimator-operator/run.sh <model>
```

This might help address the `namespace` TODO.

ehoner (changes requested):
The nits are optional. The other comments can be discussed or fixed.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/HoptimatorOperatorApp.java
ryannedolan: I'd rather keep a distinction between JAVA_OPTS and app options. I opened issue #17 to add `--properties key=value` or similar at some point.

hoptimator-cli/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,7 @@
+log4j.rootLogger=ERROR,stderr
+
+log4j.appender.stderr=org.apache.log4j.ConsoleAppender
+log4j.appender.stderr.layout=org.apache.log4j.PatternLayout
+log4j.appender.stderr.layout.ConversionPattern=%p\t%C{1}\t%m%n
```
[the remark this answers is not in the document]
ryannedolan: In the CLI it's probably just noise. I can add timestamps to the operator logs tho.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/ControllerService.java
ryannedolan: Looks like this depends on java 9, which is probably okay, but let's save that for a future PR. We can clean up a lot of small stuff like this if we skip to 11.

hoptimator-cli/src/main/resources/log4j.properties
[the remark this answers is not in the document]
ryannedolan: I dropped log4j in favor of slf4j-simple, so deleting this file.

ehoner (approved):
LGTM

Issue #16, opened 2023-05-24
!test command


ryannedolan:
We need a way to test SQL statements in hoptimator-cli (and esp in integration tests) beyond failing if they crash. Borrowing from dbt, we can consider a test as passing if it returns _zero_ results (but doesn't crash). This would allow us to write SQL that e.g. verifies that two tables contain the same data, or e.g. verifies that a column doesn't have malformed values.

Tests would look something like:

```
-- test that a table has no null values
0: hoptimator> !test SELECT X FROM DB.FOO WHERE X=NULL
```

This is tricky because most tables we encounter are unbounded streams, so we don't know when we've exhausted an entire table. If the example above involved a Kafka topic, the query would hang or timeout. In that case, we may "PASS" a test, even if subsequent malformed data arrives a second later. Moreover, from an implementation perspective, it may be difficult to distinguish between "this job crashed from some error" (FAIL) vs "this job timed-out without finding malformed values" (PASS).

An additional complexity here is that hoptimator's planner does not support LIMIT clauses. When you run something like `SELECT * FROM DB.FOO LIMIT 5` in the CLI, the LIMIT part is handled by Calcite's Enumerable calling convention -- not by Hoptimator's "pipeline" calling convention. This is because it generally makes no sense to have a pipeline with limited data. Thus, we may wish to `!test` a query that includes a LIMIT, but there will be no way to plan such a pipeline. One possible way forward here is to extend FlinkIterable with a limit, and leverage `DataStream.executeAndCollect(limit)`. Interestingly, we only need _one_ result to FAIL a test (since that would be a non-zero number of rows) so `!test` could conceivably just always use limit=1.

PR #19, opened 2023-05-27
Add !test and !checkexpect commands to Hoptimator CLI

stevenewald:
### Overview
This PR introduces two new command additions for testing purposes and also enhances the functionality of the FlinkIterable class.

## New testing commands
1. `!test SELECT ...`
Validates that the output of the SQL query contains at least one row and doesn't crash
3. `!checkexpect <value> SELECT ...`
Verifies that the output of the SQL query contains the specified value and doesn't crash

## Changes to FlinkIterable
A new overloaded field method has been added, which accepts an integer position and an Integer limit. If the limit is set to null, the method defaults to the standard field function with the provided position argument. However, if a limit is specified, it enables the CollectAndExecute function, which is particularly useful when testing streams.

The enhanced !test and !checkexpect commands now support the LIMIT keyword in SQL queries.

This PR addresses #16


stevenewald:
I worked on the publish-hoptimator-cli branch because the other branches failed to build (or failed integration tests). It looks like the merge conflicts are from merging that branch into main

ryannedolan:
> `!test` ... at least one row

I had envisioned the opposite of this -- that there are *zero* rows in the result. This is because it tends to be easier to write SQL that finds problems vs SQL that finds "at least one" non-problem. For example, say you want to check that no names are NULL. You'd want a test of the form "fail if there are any results in" `SELECT name FROM names WHERE name == NULL`. Try writing the same check using "at least one row" semantics, and it becomes really hard!

That said, clearly `!test` is not a good name for this concept, cuz we had _opposite_ ideas of what it would mean.

What if, instead, we have a family of `!check` commands, like `!check <value>`, `!check empty`, `!check not empty`? That would cover all three use-cases, with a bit extra code and a lot less confusion. wdyt?

Alternatively, we could stick with one general-purpose `!test`, but I think it'd need to mean "fail if non-zero rows".

ryannedolan:
> I worked on the publish-hoptimator-cli branch because the other branches failed to build (or failed integration tests). It looks like the merge conflicts are from merging that branch into main

The main branch is green and now has the insert-into command. I tried a trivial merge but looks like the insert-into command is coming in twice now :(. Can you rebase on main? (Or, if I were you, I'd create a new branch from main and copy over my changes there.)

stevenewald:
Good point about the !test command!

I think `!check <value>`, `!check empty`, and `!check not empty` would work, but it might be unclear what the difference is if you're just reading through a test file and are unfamiliar with the Hoptimator CLI. For example, if you showed a user the following 2 commands:
`!check sweatshirt SELECT name from INVENTORY`
`!check empty SELECT name from INVENTORY WHERE  name == NULL`

It might not be super clear that "empty" is a keyword, while "sweatshirt" is interpreted as a value.

What if we had `!checkmt`, `!checkval`, `!checknotmt`? They could all match to the same command in the code and just switch the implementation slightly based on the suffix. To use the last example,
`!checkval sweatshirt SELECT name FROM inventory`
`!checkmt SELECT name FROM inventory WHERE name == NULL`

In my mind, this is a bit more concise and makes it a easier to understand if you're reading through a test file. Happy to do either way though! Lmk what you think

Once I finish the changes I'll just branch from main and copy over.

PR #21, opened 2023-05-31
Add !check command to Hoptimator CLI

stevenewald:
Made a new PR because head repo is on main branch (vs previous publish-hoptimator-cli). The updates to FlinkIterator as mentioned in #19 are still relevant.

I updated the CLI to just have one test command (`!check`) which supports the following:
`!check empty <query>`
`!check not empty <query>`
`!check value <value> <query>`


ryannedolan (approved):
Nice!

PR #23, opened 2023-05-31
Namespace arguments for hoptimator-operator

stevenewald:
Added Apache Commons CLI to dependencies and used it to create the optional namespace argument for hoptimator-operator. While overkill for a singular flag, I'm expecting more flags will be added later, so a library for parsing arguments will make adding these much easier.

I manually tested with and without the -n/--namespace flag and it worked both times (i.e., the hoptimator-operator registered the namespace).

This PR addresses #17

### DIFF
--- a/hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/HoptimatorOperatorApp.java
+++ b/hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/HoptimatorOperatorApp.java
@@ -45,7 +45,7 @@ public class HoptimatorOperatorApp {
   }
 
   public static void main(String[] args) throws Exception {
-    if (args.length<1) {
+    if (args.length < 1) {
       throw new IllegalArgumentException("Missing model file argument.");
     }
 
--- a/hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/HoptimatorOperatorApp.java
+++ b/hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/HoptimatorOperatorApp.java
@@ -60,13 +60,13 @@ public class HoptimatorOperatorApp {
     CommandLine cmd;
 
     try {
-        cmd = parser.parse(options, args);
+      cmd = parser.parse(options, args);
     } catch (ParseException e) {
-        System.out.println(e.getMessage());
-        formatter.printHelp("utility-name", options);
+      System.out.println(e.getMessage());
+      formatter.printHelp("utility-name", options);
 
-        System.exit(1);
-        return;
+      System.exit(1);
+      return;
     }
 
     String modelFileInput = cmd.getArgs()[0];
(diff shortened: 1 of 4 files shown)


hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/HoptimatorOperatorApp.java
```diff
@@ -37,10 +45,34 @@ public HoptimatorOperatorApp(String modelPath, String namespace, Properties prop
   }
 
   public static void main(String[] args) throws Exception {
-    if (args.length != 1) {
+    if (args.length<1) {
       throw new IllegalArgumentException("Missing model file argument.");
     }
-    new HoptimatorOperatorApp(args[0], "default", new Properties()).run();
```
ryannedolan, line 43: Some formatting issues here? Looks like 4 instead of 2 spaces.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/HoptimatorOperatorApp.java
```diff
@@ -37,10 +45,34 @@ public HoptimatorOperatorApp(String modelPath, String namespace, Properties prop
   }
 
   public static void main(String[] args) throws Exception {
-    if (args.length != 1) {
+    if (args.length<1) {
```
ryannedolan, line 48:
```suggestion
    if (args.length < 1) {
```

ryannedolan (changes requested):
Nice! I'm shocked it worked o.0

Couple formatting issues, otherwise lgtm.

ehoner:
I think that the `-n` flag is good, but might be confusing with the `-nn` already in use. [hoptimator-cli/run.sh](https://github.com/linkedin/Hoptimator/blob/8b44f3bbbc5d4afa4229628be2dcfdfa32d2b39b/hoptimator-cli/run.sh#L9). Not sure which, if either, should be changed though.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/HoptimatorOperatorApp.java
stevenewald: Are you referring to line 49 or 63-69?

stevenewald:
If we're unlikely to add additional arguments besides a namespace, would it make sense for it to be an optional second argument (rather than set by a flag)? So both of the following commands would be valid:
`./hoptimator-operator <path_to_model>`
`./hoptimator-operator <path_to_model> <namespace>`

This would solve the conflicting flag issue. From the original issue #17 , it seemed like we were expecting additional arguments, but if not, I can revert the Apache Commons CLI addition and just parse the second argument.

ryannedolan:
> If we're unlikely to add additional arguments besides a namespace

No, sorry, I mean for the CLI (hoptimator-cli) the arguments are pretty much set in stone by sqlline. The operator (hoptimator-operator) has no such limitations, and may well require additional arguments down the line.

I think we should stick with what you have here. The name collision is unfortunate but understandable. Anything different from `-n, --namespace` would be even more confusing.

stevenewald:
Oh, got it! I fixed the formatting (assuming those lines were what you were referring to), should be ready to merge

ryannedolan (approved):

PR #31, opened 2023-06-07
feat: support multiple planner output dialects

HoganEdwardChu:
- Now, PipelineRel pipeline method has responsibility of dialect
- But still, we need to think a way that operator can generate pipelines in whatever dialect it wants. (i think selecting dialect can be done in different branch in this step, change method reponsibility)


ryannedolan (approved):
lgtm, thanks!

PR #33, opened 2023-06-21
Enable adapters to specify read vs write resources

### LINKED ISSUE #22: Expose whether a table is being used as a source or a sink
The Catalog API (which includes interfaces from Calcite, along with Database, HopTable, TableResolver, etc), was designed to represent data sources, and was only later leveraged for sinks as well. This happens to work because Flink Connectors tend to be symmetrical. For example, the Flink Kafka connector can be used to both read and write from Kafka using the _same_ configuration, and the _same_ table definition DDL. It's easy to write an adapter that uses such a Flink Connector, and the resulting tables can be used as either sources or sinks.

OTOH, not all Flink Connectors are symmetrical. Occasionally, the configuration (and thus, the DDL) for a Flink Connector is different when used as a source or sink. Often, a Connector supports one mode but not the other. More importantly, the `Resources` the planner generates may need to be different in either case. This means it would be valuable for an adapter to do one thing when generating resources for a source table, and another for a sink table.

In particular, we would like to add ACL support, which would involve generating Access Control Entries. Naturally, the ACE is different when a table is used as a source or a sink. Trivially, a sink table needs WRITE permission, whereas a source table only needs READ permission. If we view ACEs as a type of `Resource`, we essentially want to generate different `Resources` depending on how the table is being used.

*Some ideas*
1. We could limit this complexity to Resources *only*. This appears to neglect the asymmetrical Flink Connector problem, but technically Flink jobs _are_ `Resource`s. If we can solve this for Resource generation, we 1) have an easy way to address the ACL issue, and 2) have a hard way to address asymmetrical connectors. It's unclear how hard, but it should be possible?
2. We could add `ResourceProvider.additionalSinkResources()` with a default of empty list. This would keep the `ResourceProvider` functional interface simple, but would give a hook for implementers that need sink-specific resources.
3. We could change `HopTable` to carry around sink-specific resources in addition to the existing `Collection<Resource>`.  This could be empty unless the `ResourceProvider` has `additionalSinkResources()` overloaded (or some other mechanism provides these sink-specific resources). This would mean that any HopTable could be used as either source or sink, since the additional sink-specific resources would also be there if the planner needs them.
4. The planner would need to "implement" the additional sink-specific resources in `insertInto(sink)` etc. Basically: `sink.additionalSinkResources().forEach(x -> resource(x))`;

ryannedolan:
Resolves #22.

Added an ACL CRD and a KafkaTopic ACL controller, enabling adapters to grant read and write access to Kafka topics. The ACL CRD is purposefully abstract in order to support other authorization backends.

In order to support ACLs and similar use-cases, I've split `HopTable.resources()` into `readResources()` and `writeResources()`. This split enables the planner to implement tables differently depending on whether they are being used as sources or sinks.

Added `ResourceProvider.readResources()` and `writeResources()`, enabling adapters to dictate which resources should be used when reading vs writing.

Added builder DSL methods `ResourceProvider.readWith()`, `writeWith()`, etc, which make it easy for adapters to include read-only and write-only resources.

Added read and write ACLs to the Kafka adapter, leveraging the new read/write resources API.

Changed the syntax of the `!mermaid` and `!yaml` commands to require an `insert into` clause. These commands no longer sink to a "dummy connector", but instead require an explicit sink. This was done in order to verify that write resources were implemented correctly.

## Testing Done

```
0: hoptimator> !mermaid insert into RAWKAFKA."test-sink" SELECT * FROM RAWKAFKA."products"
```

...is rendered as:

```mermaid
flowchart
  subgraph SqlJob
  R-1886594262["sql: CREATE DATABASE I..."]
  end
  subgraph KafkaTopic
  R1216691802["clientOverrides: group.id: hoptima...
name: products
numPartitions: null"]
  R408288402["clientOverrides: group.id: hoptima...
name: test-sink
numPartitions: null"]
  end
  subgraph KafkaTopicAcl
  R1409342664["method: Read
principal: User:ANONYMOUS
topicName: products"]
  R-1832104751["method: Write
principal: User:ANONYMOUS
topicName: test-sink"]
  end
  subgraph SqlJob
    R1216691802 --> R-1886594262
  end
  subgraph KafkaTopic
    R-1886594262 --> R408288402
  end
  subgraph KafkaTopicAcl
  end
```

Notice the Read ACL for `products` and the Write ACL for `test-sink`.

In response to a corresponding Subscription, the operator successfully creates ACL rules in Kafka:

```
...
Granted User:ANONYMOUS Read access to products.
...
Granted User:ANONYMOUS Write access to test-sink.
```


hoptimator-catalog/src/main/java/com/linkedin/hoptimator/catalog/HopTable.java
```diff
@@ -27,25 +27,36 @@ public class HopTable extends AbstractTable implements ScriptImplementor, Transl
   private final String database;
   private final String name;
   private final RelDataType rowType;
-  private final Collection<Resource> resources;
+  private final Collection<Resource> readResources;
+  private final Collection<Resource> writeResources;
   private final ScriptImplementor implementor;
```
vmaheshw, line 33: This class is missing JAVADOCS.

hoptimator-kafka-adapter/src/main/java/com/linkedin/hoptimator/operator/kafka/KafkaTopicAclReconciler.java
```diff
@@ -0,0 +1,134 @@
...
+
+      // assemble AdminClient config
+      ConfigAssembler assembler = new ConfigAssembler(operator);
+      list(target.getSpec().getClientConfigs()).forEach(x ->
+        assembler.addRef(namespace, x.getConfigMapRef().getName()));
+      map(target.getSpec().getClientOverrides()).forEach((k, v) -> assembler.addOverride(k, v));
+      Properties properties = assembler.assembleProperties();
+      log.info("Using AdminClient config: {}", properties);
+
+      AdminClient admin = AdminClient.create(properties);
```
vmaheshw, line 99: Isn't this an expensive operation? Especially if the reconciliation keeps on failing for some other internal reason.

deploy/dev/kafka.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: kafka.strimzi.io/v1beta2
```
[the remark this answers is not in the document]
ryannedolan: That's a good question. I'll look into it. We aren't bundling strimzi here, so I don't believe there is any license issue. But worth checking.

deploy/samples/kafkatopics.yaml
```diff
@@ -5,5 +5,5 @@ metadata:
 spec:
   topicName: test-sink
   clientOverrides:
-    bootstrap.servers: my-cluster-kafka-bootstrap.kafka.svc:9092
+    bootstrap.servers: one-kafka-bootstrap.kafka.svc:9092
```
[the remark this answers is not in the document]
ryannedolan: I changed the name of the kafka cluster from "my-cluster" to "one". Just habit from working on MirrorMaker2 I guess :)

hoptimator-kafka-adapter/src/main/java/com/linkedin/hoptimator/operator/kafka/KafkaTopicAclReconciler.java
ryannedolan: We have a non-zero retry duration, so reconciliation shouldn't be a tight loop. But yeah, it can be sorta expensive, since it registers a bunch of metrics and whatnot.

hoptimator-catalog/src/main/java/com/linkedin/hoptimator/catalog/HopTable.java
ryannedolan: Where?

deploy/samples/acls.yaml
```diff
@@ -0,0 +1,24 @@
...
+    name: test-sink
+  method: Read
+  principal: User:ANONYMOUS
+
+---
+
+apiVersion: hoptimator.linkedin.com/v1alpha1
+kind: Acl
+metadata:
+  name: sample-kafka-acl-2
```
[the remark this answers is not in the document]
ryannedolan: We don't have any groups defined yet, so that won't work. If we did, it would look like `Group:my-group`.

deploy/dev/kafka.yaml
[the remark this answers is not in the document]
ryannedolan: I had licensed these files under Apache 2 and attributed their respective sources so we should be good.

hoptimator-cli/src/main/java/com/linkedin/hoptimator/HoptimatorCliApp.java
```diff
@@ -55,6 +55,38 @@ protected int run(String[] args) throws IOException {
     sqlline.updateCommandHandlers(commandHandlers);
     return sqlline.begin(args, null, true).ordinal();
   }
+
+  private static InsertInto parseInsertInto(String s) {
+      String sql = s.trim();
+      if (!sql.startsWith("insert into ")) {
```
hshukla, line 61: Any reason for not having case insensitive support for `insert into` but we have case insensitive support for `SELECT`?

hoptimator-kafka-adapter/src/main/java/com/linkedin/hoptimator/catalog/kafka/KafkaTopicAcl.java
```diff
@@ -0,0 +1,13 @@
+package com.linkedin.hoptimator.catalog.kafka;
+
+import com.linkedin.hoptimator.catalog.Resource;
+
+public class KafkaTopicAcl extends Resource {
```
hshukla, line 5: Can we add some javadoc on what this class is for? By reading just the constructor I am not able to understand what it actually is for.

hoptimator-kafka-adapter/src/main/java/com/linkedin/hoptimator/operator/kafka/KafkaTopicAclReconciler.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.concurrent.ExecutionException;
+
+public class KafkaTopicAclReconciler implements Reconciler {
+  private final static Logger log = LoggerFactory.getLogger(KafkaTopicAclReconciler.class);
+  private final static String ACL = "hoptimator.linkedin.com/v1alpha1/Acl";
+  private final static String KAFKATOPIC = "hoptimator.linkedin.com/v1alpha1/KafkaTopic";
```
hshukla, line 39: Nit: `KAFKA_TOPIC`

hoptimator-kafka-adapter/src/main/java/com/linkedin/hoptimator/operator/kafka/KafkaTopicAclReconciler.java
```diff
@@ -0,0 +1,134 @@
...
+  private final Operator operator;
+
+  public KafkaTopicAclReconciler(Operator operator) {
+    this.operator = operator;
+  }
+
+  @Override
+  public Result reconcile(Request request) {
+    log.info("Reconciling request {}", request);
+    String name = request.getName();
```
hshukla, line 50: Can request be null here or we do that validation before the call comes here?

hoptimator-kafka-adapter/src/main/java/com/linkedin/hoptimator/operator/kafka/KafkaTopicAclReconciler.java
```diff
@@ -0,0 +1,134 @@
...
+      AclOperation operation;
+      switch (method) {
+      case READ:
+        operation = AclOperation.READ;
+        break;
+      case WRITE:
+        operation = AclOperation.WRITE;
+        break;
+      default:
+        log.info("Unsupported KafkaTopic Acl operation {}. Skipping.", method);
```
hshukla, line 77: is this actually `info` or `warn` or `error` something?

deploy/acls.crd.yaml
```diff
@@ -0,0 +1,105 @@
...
+                  - Patch
+                  - Trace
+                  - Options
+                  - GetAll
+                  - BatchGet
+                  - BatchCreate
+                  - BatchUpdate
+                  - PartialUpdate
+                  - BatchDelete
+                  - BatchPartialDelete
```
hshukla, line 68: Do we have to expose all the options? are we planning to support or need all? Did we take this from somewhere or this is something we think we need to offer?

hoptimator-kafka-adapter/src/main/java/com/linkedin/hoptimator/operator/kafka/KafkaTopicAclReconciler.java
ryannedolan: I think it can't be null, or else this method wouldn't be invoked. There doesn't need to be an object but there must be a request of some kind... I think.

hoptimator-cli/src/main/java/com/linkedin/hoptimator/HoptimatorCliApp.java
ryannedolan: Good idea! I just cargo-culted this behavior from sqlline, but I don't see a reason to not make it case insensitive.

hoptimator-kafka-adapter/src/main/java/com/linkedin/hoptimator/operator/kafka/KafkaTopicAclReconciler.java
ryannedolan: This would be normal, e.g. if some other reconciler handled this request instead of this one. I suppose it could be WARN.

deploy/acls.crd.yaml
ryannedolan: These are all the methods we currently support internally. It's pretty exhaustive but could always be expanded as needed.

hoptimator-kafka-adapter/src/main/java/com/linkedin/hoptimator/catalog/kafka/KafkaTopicAcl.java
ryannedolan: I'll mark these private. They aren't really part of the API.

hshukla (approved):

Issue #36, opened 2023-06-22
CLI logs are too verbose


ryannedolan:
The CLI and operator currently use `slf4j-simple` with default configuration. This is too verbose for the CLI, as Flink generates a ton of logs.

HoganEdwardChu:
Hi, I didn't mean to disturb you, I was watching with interest. You mean like this https://github.com/HoganEdwardChu/Hoptimator/pull/2

HoganEdwardChu:
I agree with you! I think using system property with sl4j-simple that way is much more explicit than custom log4j configuration

PR #37, opened 2023-06-26
Subscription "hints"

ryannedolan:
Added `Subscription.spec.hints` field, which can be used to tweak _sink resources_, i.e. any resources created downstream of the SQL job (e.g. an output Kafka topic). Hints are ignored by the SQL engine.

Added `SubscriptionEnvironment`, which exposes hints to resource templates. While I was here, I also exposed the computed Avro schema and changed `name` and `namespace` to `pipeline.name` and `pipeline.namespace`. This should be less confusing than having "name" in templates, since "name" is very overloaded in that context.

Added support for default values in template expressions via`{{property : defaultValue}}`. This was required in order to support hints, since hints are usually undefined.

Changed the KafkaTopic template to now use a `numPartitions` hint. This means you can change the number of partitions in a pipeline by specifying `Subscription.spec.hints.numPartitions`. This gets carried through to the KafkaTopic controller, which will add partitions if necessary.

## Details

The idea behind "hints" is that we need to be able to control low-level aspects of the data plane, but we also don't want to surface this complexity to the SQL layer. This is a unique aspect of Hoptimator -- the SQL is very generic, and purposefully avoids anything that may be considered an "implementation detail". One way this goal manifests is that we do not support a `WITH(...)` clause, unlike, say, Flink SQL or KSQLdb. Instead, we can give "hints" to the control plane out-of-band in the Subscription object.

As the name implies, hints are not specifications. The underlying controllers may ignore hints, especially if they are impossible to satisfy.

## Testing Done

To verify the new default values mechanism, we can have the CLI generate YAML for a pipeline:

```
> !yaml insert into RAWKAFKA."test-sink" SELECT AGE AS PAYLOAD, NAME AS KEY FROM DATAGEN.PERSON
...

kind: FlinkDeployment
metadata:
  name: {{pipeline.name}}-flink-job
  namespace: {{pipeline.namespace}}
spec:
--->%---
    taskmanager.numberOfTaskSlots: "1"
  serviceAccount: flink
  jobManager:
    resource:
      memory: "2048m"
      cpu: .1
  taskManager:
    resource:
      memory: "2048m"
      cpu: .1
--->%---
kind: KafkaTopic
metadata:
  name: {{pipeline.name}}-kafka-topic-9d0eb3f4
  namespace: {{pipeline.namespace}}
spec:
  topicName: test-sink
  numPartitions: null
--->%---
```

As expected, `pipeline.name` and `pipeline.namespace` are not expanded, as they are missing from the environment. (There is no Subscription being created here.)

OTOH, we see the operator fills these in:

```
--->%---
kind: FlinkDeployment
metadata:
  name: products-flink-job
  namespace: default
--->%---
```

If we deploy a Subscription with hints:

```
kind: Subscription
metadata:
  name: products
spec:
  sql: SELECT "quantity", "product_id" AS KEY FROM INVENTORY."products_on_hand"
  database: RAWKAFKA
  hints:
    numPartitions: "2".  # <--- added hint here
```

... we see the operator generates appropriate YAML:

```
kind: KafkaTopic
metadata:
  name: products-kafka-topic-477cf338
  namespace: default
spec:
  topicName: products
  numPartitions: 2
--->%---
```

Most importantly, we see the KafkaTopic controller reacts to the change in `numPartitions`:

```
Desired partitions 2 > actual partitions 1. Creating additional partitions.
```


deploy/samples/subscriptions.yaml
```diff
@@ -6,4 +6,5 @@ metadata:
 spec:
   sql: SELECT "quantity", "product_id" AS KEY FROM INVENTORY."products_on_hand"
   database: RAWKAFKA
-
+  hints:
+    kafka.numPartitions: "2"
```
vmaheshw, line 10: It is unclear whether this hint is for the Hoptimator to get more information about the source, or it is a hint towards the destination. This is important especially, that the User is submitting the select * query, without any mention of the destination/destination type. If this information is hard-coded, or inferred from hints, then hints keywords should be more contextual.

hoptimator-catalog/src/main/java/com/linkedin/hoptimator/catalog/Resource.java
```diff
@@ -182,31 +186,38 @@ public SimpleEnvironment with(String key, String value) {
...
     @Override
-    public String get(String key) {
+    public String getOrDefault(String key, String defaultValue) {
       String value = System.getenv(key);
       if (value == null) {
         value = System.getProperty(key);
       }
+      if (value == null) {
+        value = defaultValue;
+      }
```
vmaheshw, line 220: Can you refactor this to `String.getOrDefault`?

hoptimator-cli/run.sh
```diff
@@ -5,5 +5,6 @@ java \
   --add-opens java.base/java.util=ALL-UNNAMED \
   --add-opens java.base/java.time=ALL-UNNAMED \
   -classpath "/opt/plugins/*/lib/*:./hoptimator-cli-all.jar" \
+  -Dorg.slf4j.simpleLogger.defaultLogLevel=error \
```
vmaheshw, line 8: should `error` be in UPPER-CASE `ERROR`?

How to override to other levels for debugging?

hoptimator-kafka-adapter/src/main/java/com/linkedin/hoptimator/catalog/kafka/KafkaTopic.java
```diff
@@ -7,12 +7,9 @@
 import java.util.Map;
 
 public class KafkaTopic extends Resource {
```
vmaheshw, line 9: This class is missing JAVADOCS.

hoptimator-kafka-adapter/src/main/java/com/linkedin/hoptimator/catalog/kafka/KafkaTopicAcl.java
```diff
@@ -3,9 +3,9 @@
 import com.linkedin.hoptimator.catalog.Resource;
 
 public class KafkaTopicAcl extends Resource {
```
vmaheshw, line 5: JAVADOCS missing.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/subscription/SubscriptionReconciler.java
```diff
@@ -65,13 +62,17 @@ public Result reconcile(Request request) {
         return new Result(false);
       }
 
+
```
vmaheshw, line 65: nit: empty line

deploy/samples/subscriptions.yaml
ryannedolan: The Kafka topic controller will try to apply `kafka.numPartitions` to all Kafka topics used by a pipeline, since the hint is the same throughout the entire pipeline. This means an existing Kafka topic could get re-partitioned even if it is only used as a source. This sorta makes sense, since end-to-end pipelines probably _should_ be partitioned all the same. And if the controller cannot honor a hint, it will just give up. Repartitioning an existing Kafka topic could just fail and we move on.

My proposed logic is: we try to honor hints across the entire pipeline. For source topics that we don't own, we won't have permission to add partitions, which is fine. For source topics that we do own (e.g. topics that are sinks in some other subscription), we can go ahead and add partitions. That's probably the least-confusing behavior.

With #33, it's possible for adapters to have different templates (and thus, different hints) for reading vs writing. This gives us a way around the default behavior for cases where it doesn't make sense. For example, we could ignore `numPartitions` when reading from a Kafka topic, and enforce it when writing to a Kafka topic.

hoptimator-cli/run.sh
ryannedolan: Lowercase is correct. For specific classes, you can use `org.slf4j.simpleLogger.log.a.b.c=debug` etc. See https://www.slf4j.org/api/org/slf4j/simple/SimpleLogger.html.

Also see #36. @HoganEdwardChu proposed switching to log4j. I don't have a strong preference, but I don't think log4j is worth the complexity.

hoptimator-kafka-adapter/src/main/java/com/linkedin/hoptimator/catalog/kafka/KafkaTopicAcl.java
ryannedolan: Maybe these should be private?

deploy/subscriptions.crd.yaml
```diff
@@ -38,6 +38,11 @@ spec:
                 database:
                   description: The database in which to create the output/sink table.
                   type: string
+                hints:
+                  description: Hints to adapters, which may disregard them.
```
hshukla, line 42: Nit: it would be good to more explicit on why it may disregard on hints.

hoptimator-catalog/src/main/java/com/linkedin/hoptimator/catalog/Resource.java
```diff
@@ -182,31 +186,38 @@ public SimpleEnvironment with(String key, String value) {
     }
 
     @Override
-    public String get(String key) {
-      if (!vars.containsKey(key)) {
+    public String getOrDefault(String key, String defaultValue) {
+      if (defaultValue == null && !vars.containsKey(key)) {
```
hshukla, line 190: what happens if the `defaultValue` is an empty or whitespace only string?

ryannedolan: It should get rendered by the template as a blank, which may actually be what you want in some cases, I guess.

I supposed you could even use the string `"null"`, e.g. you could have the hint `subscription.spec.hints.kafka.numPartitions: "null"` s.t. the template renders `numPartitions: null`, which the Kafka controller interprets as "whatever the broker has configured".

Notice hints are just strings, so we should accept `""`, `"null"`, or whatever the user wants to inject into the template.

ryannedolan: (...and the same with `defaultValue` ... should be whatever raw string the template provides.)

hshukla: thinking more, what if someone wants to provide more metadata in hints? don't we need to make it object? do we need to namespace by the adapter type? like if you say database is queueing , hints object needs to have Kafka/stream related metadata Vs database=brooklin, hints need to have Brooklin datastream related hints?

ryannedolan: Yes, I wonder about this too. Instead of `kafka.numPartitions` we could have e.g. `QUEUING.numPartitions`, which would solve some problems I mentioned above to @vmaheshw. Specifically, the current implementation here applies to _all_ Kafka topics in a pipeline, which _probably_ isn't what you want. If you have a pipeline that reads from a Kafka topic and writes to another, the current implementation will try to add partitions to the sink topic _and the source topic_, which is a bit weird.

If instead we had e.g. `QUEUING.numPartitions` vs `TRACKING.numPartitions`, then we could explicitly control the partitions for the source and sink independently _iff_ they happen to be in different databases. If the same database is used as both source and sink (e.g. QUEUING --> QUEUING), this doesn't really help us! So I don't think using the database name is a perfect solution either.

We could maybe have hints like `QUEUING.my-topic.numPartitions`, i.e. we could scope hints down to the specific table in a specific database. That would solve the problems above, but it might be hard to provide hints to internal tables. For example, if we have a CDC stream as part of the pipeline, we'd maybe have something like:

```
spec:
   sql: SELECT * FROM MYSQL.MYTABLE
   hints:
     CDC.MYTABLE_CHANGES.numPartitions: "2"
```

... which works fine but means that the user must know about the internal table `CDC.MYTABLE_CHANGES`.

I like your idea of making these objects instead of just strings. What about smth like:

````
hints:
   - database: QUEUING
      table: my_topic
      key: numPartitions
      value: "4"
```

... where the database and table properties are optional.

If that makes sense, lets add support for the global scope (minus database and table scope) for now, and it will be easy to add the database and table scope later.

ryannedolan: We talked out-of-band and decided that the most common (and really, the only logical) use-case for hints is to control resources related to _sinks_. So instead of enabling hints to be scoped to arbitrary resources, we decided to limit them just to sinks, i.e. all the resources related to the sink table.

hshukla (approved):

PR #38, opened 2023-06-28
Report subscription status

ryannedolan:
Added logic to the subscription reconciler to keep track of deployed resources and to periodically check on their readiness. Subscription readiness is now predicated on the readiness of all resources in the pipeline.

## Testing

Subscription status now progresses through multiple states:

```
$ kubectl get subscriptions -w
NAME       STATUS      DB         SQL
products               RAWKAFKA   SELECT "quantity", "product_id" AS KEY FROM INVENTORY."products_on_hand"
products   Planned.    RAWKAFKA   SELECT "quantity", "product_id" AS KEY FROM INVENTORY."products_on_hand"
products   Deployed.   RAWKAFKA   SELECT "quantity", "product_id" AS KEY FROM INVENTORY."products_on_hand"
products   Ready.      RAWKAFKA   SELECT "quantity", "product_id" AS KEY FROM INVENTORY."products_on_hand"
```

While the pipeline is being deployed, the operator periodically checks on the status of all resources:

```
Checking status of pipeline for Subscription/products...
Resource default/KafkaTopic/products-kafka-topic considered ready by default.
KafkaTopic/products-kafka-topic is ready.
FlinkDeployment/products-flink-job is NOT ready.
Pipeline for Subscription/products is NOT ready.
...
FlinkDeployment/products-flink-job is ready.
Subscription/products is ready.
```

### DIFF
--- a/deploy/subscriptions.crd.yaml
+++ b/deploy/subscriptions.crd.yaml
@@ -51,9 +51,21 @@ spec:
                 message:
                   description: Error or success message, for information only.
                   type: string
+                sql:
+                  description: The SQL being implemented by this pipeline.
+                  type: string
+                resources:
+                  description: The resources in the pipeline.
+                  type: array
+                  items:
+                    type: string
       subresources:
         status: {}
       additionalPrinterColumns:
+      - name: STATUS
+        type: string
+        description: Status message from the operator.
+        jsonPath: .status.message
       - name: DB
         type: string
         description: The database where the subscription is materialized.
--- a/deploy/subscriptions.crd.yaml
+++ b/deploy/subscriptions.crd.yaml
@@ -55,7 +55,7 @@ spec:
                   description: The SQL being implemented by this pipeline.
                   type: string
                 resources:
-                  description: The resources in the pipeline.
+                  description: The YAML generated to implement this pipeline.
                   type: array
                   items:
                     type: string
--- a/hoptimator-models/src/main/java/com/linkedin/hoptimator/models/V1alpha1KafkaTopic.java
+++ b/hoptimator-models/src/main/java/com/linkedin/hoptimator/models/V1alpha1KafkaTopic.java
@@ -31,7 +31,7 @@ import java.io.IOException;
  * Kafka Topic
  */
 @ApiModel(description = "Kafka Topic")
-@javax.annotation.Generated(value = "org.openapitools.codegen.languages.JavaClientCodegen", date = "2023-06-28T05:08:08.646Z[Etc/UTC]")
+@javax.annotation.Generated(value = "org.openapitools.codegen.languages.JavaClientCodegen", date = "2023-06-29T22:51:02.615Z[Etc/UTC]")
 public class V1alpha1KafkaTopic implements io.kubernetes.client.common.KubernetesObject {
   public static final String SERIALIZED_NAME_API_VERSION = "apiVersion";
   @SerializedName(SERIALIZED_NAME_API_VERSION)
--- a/hoptimator-models/src/main/java/com/linkedin/hoptimator/models/V1alpha1KafkaTopic.java
+++ b/hoptimator-models/src/main/java/com/linkedin/hoptimator/models/V1alpha1KafkaTopic.java
@@ -31,7 +31,7 @@ import java.io.IOException;
  * Kafka Topic
  */
 @ApiModel(description = "Kafka Topic")
-@javax.annotation.Generated(value = "org.openapitools.codegen.languages.JavaClientCodegen", date = "2023-04-28T19:46:31.976Z[Etc/UTC]")
+@javax.annotation.Generated(value = "org.openapitools.codegen.languages.JavaClientCodegen", date = "2023-06-28T05:08:08.646Z[Etc/UTC]")
 public class V1alpha1KafkaTopic implements io.kubernetes.client.common.KubernetesObject {
   public static final String SERIALIZED_NAME_API_VERSION = "apiVersion";
   @SerializedName(SERIALIZED_NAME_API_VERSION)
--- a/hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/subscription/SubscriptionReconciler.java
+++ b/hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/subscription/SubscriptionReconciler.java
@@ -3,8 +3,8 @@ package com.linkedin.hoptimator.operator.subscription;
 import com.linkedin.hoptimator.catalog.Resource;
 import com.linkedin.hoptimator.catalog.HopTable;
 import com.linkedin.hoptimator.models.V1alpha1Subscription;
-import com.linkedin.hoptimator.models.V1alpha1SubscriptionStatus;
 import com.linkedin.hoptimator.models.V1alpha1SubscriptionSpec;
+import com.linkedin.hoptimator.models.V1alpha1SubscriptionStatus;
 import com.linkedin.hoptimator.operator.Operator;
 import com.linkedin.hoptimator.operator.ConfigAssembler;
 import com.linkedin.hoptimator.operator.RequestEnvironment;
@@ -27,6 +27,7 @@ import io.kubernetes.client.util.generic.dynamic.Dynamics;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
@@ -35,6 +36,7 @@ import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.CountDownLatch;
+import java.util.stream.Collectors;
 
 public class SubscriptionReconciler implements Reconciler {
   private final static Logger log = LoggerFactory.getLogger(SubscriptionReconciler.class);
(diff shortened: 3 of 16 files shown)


hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/subscription/SubscriptionReconciler.java
```diff
@@ -106,6 +150,102 @@ Pipeline pipeline(V1alpha1Subscription object) throws Exception {
...
+    try {
+      return obj.getRaw().get("status").getAsJsonObject().get("state").getAsString()
+        .matches("(?i)READY|RUNNING|FINISHED");
+    } catch (Exception e) {
+      log.debug("Exception looking for .status.state. Swallowing.", e);
+    }
+    try {
+      return obj.getRaw().get("status").getAsJsonObject().get("jobStatus").getAsJsonObject()
+        .get("state").getAsString().matches("(?i)READY|RUNNING|FINISHED");
+    } catch (Exception e) {
```
hshukla, line 240: [Minor]: If I understand this correctly, all these different try catch blocks are so that we can have a custom log? Would it be possible to unify them and since you are anyway logging exception object `e` we will have idea on what is causing this, especially where we are catching only Exception, there is no particular runtime exception type we are catching which needs separate catch block.

deploy/subscriptions.crd.yaml
```diff
@@ -51,9 +51,21 @@ spec:
                 message:
                   description: Error or success message, for information only.
                   type: string
+                sql:
+                  description: The SQL being implemented by this pipeline.
+                  type: string
+                resources:
```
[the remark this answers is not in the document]
ryannedolan: Good call. I'll improve the description here.

The resources array is just the YAML generated by the planner. Here's an example:

```
$ kubectl get subscriptions products -o yaml
kind: Subscription
--->%---
spec:
  database: RAWKAFKA
  sql: SELECT "quantity", "product_id" AS KEY FROM INVENTORY."products_on_hand"
status:
  message: Deployed.
  ready: false
  resources:
  - |
    kind: KafkaTopic
    metadata:
      name: products-kafka-topic
      namespace: default
    spec:
      topicName: products
      numPartitions: null
      clientOverrides:
        group.id: hoptimator-test
        bootstrap.servers: my-cluster-kafka-bootstrap.kafka.svc:9092
        auto.offset.reset: earliest
  - |+
    kind: FlinkDeployment
    metadata:
      namespace: default
      name: products-flink-job
    spec:
--->%---
  sql: SELECT "quantity", "product_id" AS KEY FROM INVENTORY."products_on_hand"
```

hoptimator-models/src/main/java/com/linkedin/hoptimator/models/V1alpha1KafkaTopic.java
```diff
@@ -31,7 +31,7 @@
  * Kafka Topic
  */
 @ApiModel(description = "Kafka Topic")
-@javax.annotation.Generated(value = "org.openapitools.codegen.languages.JavaClientCodegen", date = "2023-04-28T19:46:31.976Z[Etc/UTC]")
+@javax.annotation.Generated(value = "org.openapitools.codegen.languages.JavaClientCodegen", date = "2023-06-28T05:08:08.646Z[Etc/UTC]")
```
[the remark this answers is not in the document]
ryannedolan: No idea. It's auto-generated by openAPI codegen.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/subscription/SubscriptionReconciler.java
ryannedolan: The idea here is to try a bunch of ways to check the status of a generic resource, which is difficult since there is no conventional way to report readiness. The multiple try blocks is so we don't need a bunch of `if (contains(x))` statements.

hshukla: Ok, so each downstream resource has their own format. would it be a good idea to namespace them, like if resource is x -> use this jsonPath? 

Can we make the matcher string `"(?i)READY|RUNNING|FINISHED"` constant?

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/subscription/SubscriptionReconciler.java
```diff
@@ -106,6 +150,102 @@ Pipeline pipeline(V1alpha1Subscription object) throws Exception {
...
+    try {
+      return obj.getRaw().get("status").getAsJsonObject().get("ready").getAsBoolean();
+    } catch (Exception e) {
+      log.debug("Exception looking for .status.ready. Swallowing.", e);
+    }
+    try {
+      return obj.getRaw().get("status").getAsJsonObject().get("state").getAsString()
+        .matches("(?i)READY|RUNNING|FINISHED");
+    } catch (Exception e) {
+      log.debug("Exception looking for .status.state. Swallowing.", e);
```
hshukla, line 235: should not this be at error level? Also, by looking this log string we don't have a way to know which sub resource it failed for, how can one find out that detail?

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/subscription/SubscriptionReconciler.java
```diff
@@ -106,6 +150,102 @@ Pipeline pipeline(V1alpha1Subscription object) throws Exception {
...
+    }
+    try {
+      return obj.getRaw().get("status").getAsJsonObject().get("jobStatus").getAsJsonObject()
+        .get("state").getAsString().matches("(?i)READY|RUNNING|FINISHED");
+    } catch (Exception e) {
+      log.debug("Exception looking for .status.jobStatus.state. Swallowing.", e);
+    }
+    // TODO: Look for common Conditions
+    log.warn("Resource {}/{}/{} considered ready by default.", obj.getMetadata().getNamespace(),
```
hshukla, line 244: should this be warn or info?

hshukla: [Minor] general comment, I see log have just plain value, would it be possible to have key={value} format? so if one knows the key they can search for that? Here, it seems like I need to know the value, may be we do know that, however, it would be more readable if it's in key={value} format.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/subscription/SubscriptionReconciler.java
```diff
@@ -56,38 +58,80 @@ public Result reconcile(Request request) {
...
+          status.setReady(false);
+          status.setMessage("Deployed.");
+        } else {
+          return new Result(true, operator.failureRetryDuration());
+        }
+      } else {
+        log.info("Checking status of pipeline for {}/{}...", kind, name);
+
+        boolean ready = status.getResources().stream()
+          .map(x -> checkStatus(x)).allMatch(x -> x);
```
ehoner, line 114: nit (optional): `.map(...)` is unnecessary and is equivalent to "just" `allMatch(...)`:

```java
.allMatch(this::checkStatus);
```

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/subscription/SubscriptionReconciler.java
```diff
@@ -106,6 +150,102 @@ Pipeline pipeline(V1alpha1Subscription object) throws Exception {
...
+      String resourceVersion = existing.getObject().getMetadata().getResourceVersion();
+      log.info("Updating existing downstream resource {}/{} {} as \n{}",
+        namespace, name, resourceVersion, yaml);
+      List<V1OwnerReference> owners = existing.getObject().getMetadata().getOwnerReferences();
+      if (owners == null) {
+        owners = new ArrayList<>();
+      }
+      if (owners.stream().anyMatch(x -> x.getUid().equals(ownerReference.getUid()))) {
+        log.info("Existing downstream resource {}/{} is already owned by {}/{}.",
+          namespace, name, ownerReference.getKind(), ownerReference.getName());
```
ehoner, line 174: If the owners are matched, is the subsequent API call still necessary? I was wondering if it is reasonable to `return true` here. If not here, should there be a test to see if the update is necessary, ie. `allMatch()` vs `anyMatch()`?

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/subscription/SubscriptionReconciler.java
ryannedolan: It's warn cuz it could indicate a problem. We're reporting success just on the basis that 1) the resource exists, and 2) the status didn't look like one we can interpret. Ideally we'd update this code anytime we see a warning here.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/subscription/SubscriptionReconciler.java
ryannedolan: The api call is still necessary, even if the ownership doesn't change, cuz anything else in the spec may change.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/subscription/SubscriptionReconciler.java
ryannedolan: The exception is supposed to happen here. It's not an error.

ehoner (approved):

PR #64, opened 2024-04-01
Use implicit row constructors. Optionally skip nested rows. Nullability bug fixes.

ryannedolan:
# Summary

`TableResolver` and `DataType.Struct` can now optionally skip nested `Row`s. This is useful for pipelines that can't handle complex record types.

`ScriptImplementor` will now replace explicit `Row(...)` constructors with implicit `(...)` row constructors, which is better supported by Flink before 1.18.

Added support for Avro arrays.

Small bug fixes related to nullability of Avro fields.

# Details

Added `dropNestedRows` method to `DataType.Struct`. Added `mapStruct` method to `TableResolver` to enable applying this and similar data type transformations to rows. Adapters can use this as an easy way to deal with complex schemas.

Flink SQL prior to 1.18 doesn't support arbitrary expressions inside an explicit row constructor, so we now rewrite these using an implicit row constructor. This works when the number of elements in the row is more than one.

Added support for Avro arrays, and added a commented-out impl of Avro maps. Maps aren't un-parseable in this version of Calcite, so Avro maps will need to wait for a version bump.

# Testing

New unit tests covering:

- avro --> rel and rel --> type conversions with nested rows
- nullability of nested avro rows
- drop nested rows

Also, spot-tested in a Kafka-->Kafka pipeline.

### DIFF
--- a/hoptimator-catalog/src/main/java/com/linkedin/hoptimator/catalog/AvroConverter.java
+++ b/hoptimator-catalog/src/main/java/com/linkedin/hoptimator/catalog/AvroConverter.java
@@ -27,7 +27,8 @@ public final class AvroConverter {
         .filter(x -> !x.getName().startsWith("__")) // don't write out hidden fields
         .map(x -> new Schema.Field(sanitize(x.getName()), avro(namespace, x.getName(), x.getType()), describe(x), null))
         .collect(Collectors.toList());
-      return Schema.createRecord(sanitize(name), dataType.toString(), namespace, false, fields);
+      return createAvroSchemaWithNullability(Schema.createRecord(sanitize(name), dataType.toString(), namespace, false, fields),
+        dataType.isNullable());
     } else {
       switch (dataType.getSqlTypeName()) {
       case INTEGER:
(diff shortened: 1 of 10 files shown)


hoptimator-catalog/src/main/java/com/linkedin/hoptimator/catalog/ScriptImplementor.java
```diff
@@ -1,23 +1,32 @@
...
 
+import org.apache.calcite.rel.RelNode;
+import org.apache.calcite.rel.type.RelDataType;
+import org.apache.calcite.rel.type.RelDataTypeFactory;
+import org.apache.calcite.rel.type.RelDataTypeField;
+import org.apache.calcite.rel.type.RelDataTypeSystem;
+import org.apache.calcite.rel.rel2sql.RelToSqlConverter;
+import org.apache.calcite.rel.rel2sql.SqlImplementor;
 import org.apache.calcite.sql.SqlWriter;
 //import org.apache.calcite.sql.SqlWriterConfig;
```
hshukla, line 11: can this be removed?

hoptimator-catalog/src/main/java/com/linkedin/hoptimator/catalog/AvroConverter.java
```diff
@@ -74,17 +87,24 @@ public static RelDataType rel(Schema schema, RelDataTypeFactory typeFactory) {
...
-      return createRelTypeWithNullability(typeFactory, SqlTypeName.FLOAT, schema.isNullable());
+      return createRelType(typeFactory, SqlTypeName.FLOAT);
     case DOUBLE:
-      return createRelTypeWithNullability(typeFactory, SqlTypeName.DOUBLE, schema.isNullable());
+      return createRelType(typeFactory, SqlTypeName.DOUBLE);
+    case BOOLEAN:
+      return createRelType(typeFactory, SqlTypeName.BOOLEAN);
+    case ARRAY:
+      return typeFactory.createArrayType(rel(schema.getElementType(), typeFactory), -1);
+//  TODO support map types
```
hshukla, line 105: any reason we are not enabling this.

hoptimator-catalog/src/main/java/com/linkedin/hoptimator/catalog/ScriptImplementor.java
ryannedolan: I'll add a comment. Leaving it in cuz it will be needed in next Calcite version.

hoptimator-catalog/src/main/java/com/linkedin/hoptimator/catalog/AvroConverter.java
ryannedolan: Need a Calcite version bump first. Map types don't seem to be unparseable in our current version of Calcite. I'll fix the comment. I'll follow up with a version bump at some point.

hshukla (approved):

PR #78, opened 2024-12-17
Add Venice support

jogrogan:
Adds Venice support to Hoptimator.
- Queries Venice for Key & Value Avro schemas, merges them into one schema, keys structured as `KEY$` to prevent collisions
- Passes `KEY` fields through to Sink options (intended to be used by flink under key.fields connector property)
- Supports partial inserts of the form `insert into "VENICE-CLUSTER0"."test-store-1" ("KEY$id", "stringField") SELECT ...`
  - Fields not specified in the insert are required to be nullable (this may still be a problem for fields that are not nullable but do have a default in Avro)
  - Captures the targetFields specified in the insert, and drops other fields.
  - Rewrites INSERT with aliasing for fields specified (see examples in comments around ScriptImplementor)

Other changes in this PR:
- Adds Venice tests
- Adds ability to stand up Venice locally in Docker
- Clean up Makefile targets
- Address various checkstyle errors

Implemented the Venice driver/schema classes with separate overridable functions to be able to handle company-internal connection components via a simple override

```
./hoptimator
0: Hoptimator> create or replace materialized view "VENICE-CLUSTER0"."my-store" as select * from "VENICE-CLUSTER0"."test-store";

0: Hoptimator> !tables
+-----------+-----------------+------------------+-------------------+---------+
| TABLE_CAT |   TABLE_SCHEM   |    TABLE_NAME    |    TABLE_TYPE     | REMARKS |
+-----------+-----------------+------------------+-------------------+---------+
...
|           | VENICE-CLUSTER0 | my-store         | MATERIALIZED VIEW |         |
|           | VENICE-CLUSTER0 | test-store       | TABLE             |         |
|           | VENICE-CLUSTER0 | test-store-1     | TABLE             |         |
...
+-----------+-----------------+------------------+-------------------+---------+
```

```
$ k get flinkdeployments.flink.apache.org -o yaml venice-cluster0-my-store
kind: FlinkDeployment
metadata:
  creationTimestamp: "2024-12-24T20:38:41Z"
  finalizers:
  - flinkdeployments.flink.apache.org/finalizer
  generation: 2
  name: venice-cluster0-my-store
  namespace: default
  resourceVersion: "32186"
  uid: e0d855d3-af15-49d9-a16b-13c87579a23e
spec:
  flinkConfiguration:
    taskmanager.numberOfTaskSlots: "1"
  flinkVersion: v1_16
  image: docker.io/library/hoptimator-flink-runner
  imagePullPolicy: Never
  job:
    args:
    - CREATE TABLE IF NOT EXISTS `test-store` (`intField` INTEGER, `stringField` VARCHAR,
      `KEY` ROW(`id` INTEGER)) WITH ('connector'='venice', 'key.fields'='KEY_id',
      'key.fields-prefix'='KEY_', 'partial-update-mode'='true', 'storeName'='test-store',
      'value.fields-include'='EXCEPT_KEY');
    - CREATE TABLE IF NOT EXISTS `my-store` (`intField` INTEGER, `stringField` VARCHAR,
      `KEY_id` INTEGER) WITH ('connector'='venice', 'key.fields'='KEY_id', 'key.fields-prefix'='KEY_',
      'partial-update-mode'='true', 'storeName'='my-store', 'value.fields-include'='EXCEPT_KEY');
    - INSERT INTO `my-store` (`intField`, `stringField`, `KEY_id`) SELECT * FROM `VENICE-CLUSTER0`.`test-store`;
    entryClass: com.linkedin.hoptimator.flink.runner.FlinkRunner
    jarURI: local:///opt/hoptimator-flink-runner.jar
    parallelism: 1
    state: running
    upgradeMode: stateless
  jobManager:
    replicas: 1
    resource:
      cpu: 0.1
      memory: 2048m
  serviceAccount: flink
  taskManager:
    resource:
      cpu: 0.1
      memory: 2048m
status:
  clusterInfo: {}
  jobManagerDeploymentStatus: DEPLOYING
  jobStatus:
    checkpointInfo:
      lastPeriodicCheckpointTimestamp: 0
    jobId: 49de826b428ba2d93e4fac8b120de2c4
    savepointInfo:
      lastPeriodicSavepointTimestamp: 0
      savepointHistory: []
    state: RECONCILING
  lifecycleState: DEPLOYED
  observedGeneration: 2
  reconciliationStatus:
    lastReconciledSpec: '{"spec":{"job":{"jarURI":"local:///opt/hoptimator-flink-runner.jar","parallelism":1,"entryClass":"com.linkedin.hoptimator.flink.runner.FlinkRunner","args":["CREATE
      TABLE IF NOT EXISTS `test-store` (`intField` INTEGER, `stringField` VARCHAR,
      `KEY` ROW(`id` INTEGER)) WITH (''connector''=''venice'', ''key.fields''=''KEY_id'',
      ''key.fields-prefix''=''KEY_'', ''partial-update-mode''=''true'', ''storeName''=''test-store'',
      ''value.fields-include''=''EXCEPT_KEY'');","CREATE TABLE IF NOT EXISTS `my-store`
      (`intField` INTEGER, `stringField` VARCHAR, `KEY_id` INTEGER) WITH (''connector''=''venice'',
      ''key.fields''=''KEY_id'', ''key.fields-prefix''=''KEY_'', ''partial-update-mode''=''true'',
      ''storeName''=''my-store'', ''value.fields-include''=''EXCEPT_KEY'');","INSERT
      INTO `my-store` (`intField`, `stringField`, `KEY_id`) SELECT * FROM `VENICE-CLUSTER0`.`test-store`;"],"state":"running","savepointTriggerNonce":null,"initialSavepointPath":null,"checkpointTriggerNonce":null,"upgradeMode":"stateless","allowNonRestoredState":null,"savepointRedeployNonce":null},"restartNonce":null,"flinkConfiguration":{"taskmanager.numberOfTaskSlots":"1"},"image":"docker.io/library/hoptimator-flink-runner","imagePullPolicy":"Never","serviceAccount":"flink","flinkVersion":"v1_16","ingress":null,"podTemplate":null,"jobManager":{"resource":{"cpu":0.1,"memory":"2048m","ephemeralStorage":null},"replicas":1,"podTemplate":null},"taskManager":{"resource":{"cpu":0.1,"memory":"2048m","ephemeralStorage":null},"replicas":null,"podTemplate":null},"logConfiguration":null,"mode":null},"resource_metadata":{"apiVersion":"flink.apache.org/v1beta1","metadata":{"generation":2},"firstDeployment":true}}'
    reconciliationTimestamp: 1735072722874
    state: DEPLOYED
  taskManager:
    labelSelector: component=taskmanager,app=venice-cluster0-my-store
    replicas: 1
```

See included tests for more samples


hoptimator-kafka/src/test/resources/kafka-ddl.id
```diff
@@ -24,8 +24,8 @@ spec:
   job:
     entryClass: com.linkedin.hoptimator.flink.runner.FlinkRunner
     args:
-    - CREATE TABLE IF NOT EXISTS `existing-topic-2` (`KEY` VARCHAR, `VALUE` BINARY) WITH ('properties.bootstrap.servers'='localhost:9092', 'topic'='existing-topic-2', 'connector'='kafka')
-    - CREATE TABLE IF NOT EXISTS `existing-topic-1` (`KEY` VARCHAR, `VALUE` BINARY) WITH ('properties.bootstrap.servers'='localhost:9092', 'topic'='existing-topic-1', 'connector'='kafka')
+    - CREATE TABLE IF NOT EXISTS `existing-topic-2` (`KEY` VARCHAR, `VALUE` BINARY) WITH ('properties.bootstrap.servers'='localhost:9092', 'connector'='kafka', 'topic'='existing-topic-2')
+    - CREATE TABLE IF NOT EXISTS `existing-topic-1` (`KEY` VARCHAR, `VALUE` BINARY) WITH ('properties.bootstrap.servers'='localhost:9092', 'connector'='kafka', 'topic'='existing-topic-1')
```
jogrogan, line 28: Not sure why this order changed, was causing tests to fail locally

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/PipelineRel.java
```diff
@@ -78,7 +82,22 @@ public void setSink(String database, List<String> path, RelDataType rowType, Map
       this.sinkDatabase = database;
       this.sinkPath = path;
       this.sinkRowType = rowType;
-      this.sinkOptions = options;
+      this.sinkOptions = addKeysAsOption(options, rowType);
+    }
+
+    private Map<String, String> addKeysAsOption(Map<String, String> options, RelDataType rowType) {
```
jogrogan, line 88: I don't love this approach, very brute force.

I looked into hints to solve this and did get them working to an extent (will open a separate PR) but this would require users to pass in key information into their SQL statement. I have not figure out a way to inject hints at runtime from VeniceDriver.

hoptimator-venice/src/main/java/com/linkedin/hoptimator/venice/VeniceStore.java
```diff
@@ -0,0 +1,62 @@
...
+  public VeniceStore(StoreSchemaFetcher storeSchemaFetcher) {
+    this.storeSchemaFetcher = storeSchemaFetcher;
+  }
+
+  @Override
+  public RelDataType getRowType(RelDataTypeFactory typeFactory) {
+    Schema keySchema = storeSchemaFetcher.getKeySchema();
+    Schema valueSchema = storeSchemaFetcher.getLatestValueSchema();
+
+    // Modify keySchema fields to contain a prefix of "KEY_" for their name in order to not clash with the value schema
```
ryannedolan, line 30: Another way to do this: have one field `KEY` which is either a simple type (if one key) or a `Row(...)` of all key fields (if more than one key). That would yield `KEY.FOO` instead of `KEY_FOO`.

Makefile
```diff
@@ -9,8 +9,8 @@ build:
 
 bounce: build undeploy deploy deploy-samples deploy-config deploy-demo
 
-# Integration tests expect K8s and Kafka to be running
-integration-tests: deploy-dev-environment deploy-samples
+# Integration tests expect K8s, Kafka, and Venice to be running
```
ryannedolan: 🔥 🔥 🔥

hoptimator-kafka/src/test/resources/kafka-ddl.id
ryannedolan: I recently added `.sorted()` to `QuidemTestBase` so this shouldn't happen going forward. These were non-deterministic previously.

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/PipelineRel.java
ryannedolan: I guess I'm surprised we need to fully specify the keys in the options. The Kafka connector has similar properties (`key.prefix`, `key.fields`), but you don't need both. Is the Venice connector doing something different here? I'd expect `key.prefix=key_` to be sufficient.

ryannedolan: Also, how would the Venice connector behave if we grouped the keys in a `Row(...)` object? Can we just have `key.fields=KEY` and then `KEY ROW(F1 VARCHAR, F2 INT)` etc?

(Not suggesting we do that, just asking if possible?)

hoptimator-kafka/src/test/resources/kafka-ddl.id
jogrogan: I reverted this and it's passing now anyway

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/PipelineRel.java
jogrogan: > I guess I'm surprised we need to fully specify the keys in the options. The Kafka connector has similar properties (`key.prefix`, `key.fields`), but you don't need both. Is the Venice connector doing something different here? I'd expect `key.prefix=key_` to be sufficient.

Yea I confirmed it is an issue with Venice due do some additional avro schema validation they do. `key.prefix` and `key.fields` do seem to be used differently and need to be used together.

jogrogan: Looking into the ROW syntax and it doesn't seem that is possible in Flink, there is no way to get Flink to destruct that ROW

ryannedolan: Makes sense. I think that's fine. The only slight concern I have is potential deviation from the Kafka -> Kafka use-case. With Kafka -> Kafka users can do `select *` and retain partitioning, since the output topic will use the input topic's `KEY` field. Users might expect `select *` to work similarly with Kafka -> Venice, except Kafka has one key (`KEY`) and Venice has one or more keys (`KEY_xyz`), so that might not work. I'm thinking we should actually change the way Kafka works and adopt this `KEY_xzy` approach, or maybe have `PipelineRel` explicitly use `KEY` in cases where there is only one key.

deploy/docker/venice/keySchema.avsc
```diff
@@ -0,0 +1,11 @@
+{
+  "type": "record",
```
ryannedolan, line 2: We need to get `create table venice.foo` working :)

hoptimator-avro/build.gradle
```diff
@@ -1,9 +1,55 @@
...
+  id 'maven-publish'
 }
 
 dependencies {
   implementation project(':hoptimator-api')
   implementation libs.avro
   implementation libs.calcite.core
 }
+
+publishing {
```
ryannedolan, line 12: Just did the same thing :)

deploy/docker/venice/keySchema.avsc
jogrogan: Agreed, I will spend some time looking into this when I can. Should be a simple API call just as I'm doing to fetch schemas, just different than the current paradigm since it isn't managed via K8s.

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/ScriptImplementor.java
```diff
@@ -245,38 +254,93 @@ public void implement(SqlWriter w) {
...
+
+  // By default, "projectNamed" will try to provide an optimization by not creating a new project if the
+  // field types are the same. This is not desirable in the insert case as the field names need to match the sink.
+  //
+  // Example:
+  // INSERT INTO `my-store` (`KEY_id`, `stringField`) SELECT * FROM `KAFKA`.`existing-topic-1`;
+  // Without forced projection this will get optimized to:
+  // INSERT INTO `my-store` (`KEYFIELD`, `VARCHARFIELD`) SELECT * FROM `KAFKA`.`existing-topic-1`;
+  // With forced project this will resolve as:
+  // INSERT INTO `my-store` (`KEY_id`, `stringField`) SELECT `KEYFIELD` AS `KEY_id`, \
```
ryannedolan, line 316: neat!

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/ScriptImplementor.java
```diff
@@ -245,38 +254,93 @@ public void implement(SqlWriter w) {
...
+    // Drops NULL fields
+    // Drops additional fields by schema
+    private static RelNode dropFields(String schema, RelNode relNode,
+        ImmutableList<Pair<Integer, String>> targetFields) {
       List<Integer> cols = new ArrayList<>();
       int i = 0;
+      Set<String> targetFieldNames = targetFields.stream().map(x -> x.right).collect(Collectors.toSet());
       for (RelDataTypeField field : relNode.getRowType().getFieldList()) {
+        // TODO: Need a better way to dynamically modify the script implementer based on the schema
+        if (schema.startsWith("VENICE")
```
ryannedolan, line 290: Don't like this. The schema should be opaque. I don't understand why Venice is special here? Can we just always drop fields that are not in `targetFieldNames`? Or maybe only drop if `targetFieldNames` isn't null?

hoptimator-venice/src/main/java/com/linkedin/hoptimator/venice/LocalControllerClient.java
```diff
@@ -0,0 +1,26 @@
...
+    super(clusterName, discoveryUrls, sslFactory);
+  }
+
+  @Override
+  protected String discoverLeaderController() {
+    try {
+      URL controllerUrl = new URL(super.discoverLeaderController());
+      return controllerUrl.getProtocol() + "://localhost:" + controllerUrl.getPort();
+    } catch (MalformedURLException e) {
+      throw new RuntimeException(e);
```
ryannedolan, line 23: Probably could be a `IllegalArgumentException`.

hoptimator-venice/src/main/java/com/linkedin/hoptimator/venice/VeniceDriver.java
```diff
@@ -0,0 +1,70 @@
...
+import java.util.Properties;
+
+import org.apache.calcite.avatica.ConnectStringParser;
+import org.apache.calcite.avatica.DriverVersion;
+import org.apache.calcite.jdbc.CalciteConnection;
+import org.apache.calcite.jdbc.Driver;
+import org.apache.calcite.schema.SchemaPlus;
+
+
+/** JDBC driver for Venice topics. */
```
ryannedolan, line 16:
```suggestion
/** JDBC driver for Venice stores. */
```

hoptimator-venice/src/main/java/com/linkedin/hoptimator/venice/VeniceStore.java
```diff
@@ -0,0 +1,66 @@
...
```
ryannedolan, line 30: I have added a `DataTypeUtils` class in `hoptimator-util` in #79. We might want to make this logic generic and move to there at some point. Might be a common scenario.

ryannedolan (approved):
The `if (schema.startsWith("VENICE")...)` logic needs to be fixed, but I think we can accept the `TODO` and fix later.

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/ScriptImplementor.java
jogrogan: Yea good point, I don't like this either. Let me see what the behavior is like if I leave out fields for Kafka

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/PipelineRel.java
jogrogan: What does retaining partitioning mean for the Kafka -> Venice use case? It seems like it is more of a problem on the producer side. Users that expect the same partitioning behavior would have to key their Kafka topic using the same combination of keys as Venice. We did this in Brooklin by constructing the producer key as a simple string with key values separated by `_` from the source keys. Of course this isn't the same as identity partitioning but it does ensure that downstream consumer tasks read the same combo of keys.

Even for the Kafka -> Kafka use case, we aren't the ones consuming, the partitioning behavior comes from Flink right? I haven't looked into it to be fair, not sure how the behavior changes if you define `key.fields` there.

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/PipelineRel.java
```diff
@@ -36,36 +44,40 @@
...
+    private ImmutableList<Pair<Integer, String>> targetFields;
     private String sinkDatabase = "pipeline";
     private List<String> sinkPath = Arrays.asList(new String[]{"PIPELINE", "SINK"});
     private RelDataType sinkRowType = null;
     private Map<String, String> sinkOptions = Collections.emptyMap();
 
-    public void visit(RelNode node) throws SQLException {
-      if (query == null) {
-        query = node;
+    public void visit(RelNode node, ImmutableList<Pair<Integer, String>> targetFields) throws SQLException {
```
ryannedolan, line 62: targetFields probably makes more sense as a constructor parameter, since it doesn't appear to change for the life of the Implementor object. Also it is weird to see the visit() method break from the "visitor pattern". Probably doesn't matter for us, but sometimes visit() is invoked via reflection, so the signature probably shouldn't change.

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/PipelineRel.java
```diff
@@ -78,7 +90,26 @@ public void setSink(String database, List<String> path, RelDataType rowType, Map
       this.sinkDatabase = database;
       this.sinkPath = path;
       this.sinkRowType = rowType;
-      this.sinkOptions = options;
+      this.sinkOptions = addKeysAsOption(options, rowType);
+    }
+
+    private Map<String, String> addKeysAsOption(Map<String, String> options, RelDataType rowType) {
+      Map<String, String> newOptions = new HashMap<>(options);
```
ryannedolan, line 97: Probably should make these LinkedHashMaps to preserve order. The integration tests expect deterministic output.

ryannedolan (approved):

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/PipelineRel.java
ryannedolan: In the Kafka -> Kafka use case, `insert into kafka.foo select * from kafka.bar` intuitively keeps the key the same, since nothing in the SQL explicitly specifies otherwise. The effect should be that foo is a mirror of bar, with the same key and partition semantics. This is indeed the case today (internally). To the SQL engine, the input table always has an implicit `KEY` column, which gets selected by `select *`.

A typical use case for Kafka->Kafka pipelines is actually _dropping_ the key and re-partitioning as round-robin. This is done today via `SELECT *, NULL AS KEY...`. Hoptimator explicitly supports `NULL AS KEY`, since it is not really possible to express "select everything except the key" in SQL.

I'm less familiar with the Venice use case, but I want to make sure that both Kafka->Kafka and Kafka->Venice are intuitive and somewhat consistent.

jogrogan: Think I get what you are saying. The Venice connector as written today doesn't look like it supports the same implicit keys. It relies on explicit keys via the `key.fields` table option.

Kafka->Kafka does support this today, you could have something like:
Kafka topic `bar` has a payload containing two fields `id` & `val`.
Specifying a flink table option of `"key.fields": "id"` before applying `insert into kafka.foo select * from kafka.bar` will use the column `id` as the key for topic `foo` when doing Kafka->Kafka. If there is an actual key present in the topic `bar` I don't believe it'll actually get used anywhere, it'll effectively be dropped.

A simple statement like `insert into venice.foo select * from kafka.bar` will still work as long as `kafka.bar` contains all the columns Venice expects, both key & value columns.

ryannedolan: Currently `select * from kafka.bar` will always include a `KEY VARCHAR`, since internally all our Kafka topics are keyed with a simple string (or null). This new Venice integration expects a `KEY` field, which contains all the keys. Individually, both make sense. I'm just wondering if `KEY VARCHAR` and `KEY Row(A ..., B ...)` need some additional magic to work interchangeably. In particular, would Kafka->Venice end up with `key.fields = KEY_KEY`? Or can we detect that there is only one key and have `key.fields = KEY`?

ryannedolan: Actually, I think what you have here with `KEY_` as a magic prefix is going to be more robust than what we currently do for Kafka. Rather than try to shoehorn our existing simplistic Kafka convention here, let's adopt your approach for Kafka->Kafka as well. I think the only thing we'd need to do is change Kafka's magic column from `KEY` to `KEY_STRING` or something, and your code will just work. Kafka keys will just pop out as `KEY Row(STRING VARCHAR)` and the connector will see `key.fields = KEY_STRING`.

PR #80, opened 2025-01-07
Add ConfigProvider SPI and K8s implementation to load ConfigMap resources

jogrogan:
Add ConfigProvider SPI and K8s implementation to load ConfigMap resources
  - Add ConfigProvider specifically for K8s to load ConfigMap objects
  - Load these configs at the driver level where needed


hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sApiEndpoints.java
```diff
@@ -41,6 +43,9 @@ public final class K8sApiEndpoints {
   public static final K8sApiEndpoint<V1alpha1JobTemplate, V1alpha1JobTemplateList> JOB_TEMPLATES =
       new K8sApiEndpoint<>("JobTemplate", "hoptimator.linkedin.com", "v1alpha1", "jobtemplates", false,
           V1alpha1JobTemplate.class, V1alpha1JobTemplateList.class);
+  public static final K8sApiEndpoint<V1ConfigMap, V1ConfigMapList> CONFIG_MAP_TEMPLATES =
```
ryannedolan, line 46: This is a native resource, so maybe move up to the previous comment block. Also probably can just call this `CONFIG_MAP`

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/PropertyUtils.java
```diff
@@ -0,0 +1,37 @@
+package com.linkedin.hoptimator.util;
+
+import java.util.Properties;
+
+public final class PropertyUtils {
+
+  public static final String HOPTIMATOR_CONFIG_DOMAIN = "hoptimator.config";
```
ryannedolan, line 7: Instead of prefixes (e.g. `venice.foo=bar`) inside a common file (`hoptimator.config`) would it be useful to break out each driver's config into separate files?

```
venice.config:
  foo=bar
```

I'm thinking each driver could do smth like `loadConfigs("venice.config")` to get driver-specific configuration?

jogrogan: We could but I was thinking there would be a lot of overlap for certain fields between drivers (Like cluster or environment type properties).

ConfigMaps do let you nest properties, do you think something like that would be better?
Example:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: hoptimator-operator
data:
  env: "dev"
  cluster: "my-cluster"

  venice.properties: |
    venice.router=localhost:823
```

ryannedolan: I see. I don't think that's any better than prefixes.

What does "domain" mean here?

jogrogan: Just a prefix, I'm getting rid of that bit since I don't like that I am loading these into System properties right now, I will load these as needed per driver

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sConfigProvider.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.linkedin.hoptimator.ConfigProvider;
+
+
+public class K8sConfigProvider implements ConfigProvider {
+
+  private static final Logger log = LoggerFactory.getLogger(K8sConfigProvider.class);
+  public static final String HOPTIMATOR_CONFIG_MAP = "hoptimator-configmap";
+
+  // Loads top level configs and expands given properties under a file-like property key
```
ryannedolan, line 21: neat!

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/ConfigService.java
```diff
@@ -0,0 +1,31 @@
...
+
+  // A Prefix is appended to all system configs added to prevent collisions
+  public static Properties config(String... expansionFields) {
+    ServiceLoader<ConfigProvider> loader = ServiceLoader.load(ConfigProvider.class);
+    Properties properties = new Properties();
+    for (ConfigProvider provider : loader) {
+      try {
+        properties.putAll(provider.loadConfig(expansionFields));
+      } catch (Exception e) {
+        log.warn("Could not load config map for provider={}", provider, e);
```
ryannedolan, line 26: This should probably throw? If a provider fails to load a config, I doubt the caller would be able to continue without it.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sConfigProvider.java
```diff
@@ -0,0 +1,44 @@
...
+  public Properties loadConfig(String... expansionFields) throws Exception {
+    Map<String, String> topLevelConfigs = loadTopLevelConfig(HOPTIMATOR_CONFIG_MAP);
+    Properties p = new Properties();
+    p.putAll(topLevelConfigs);
+    for (String expansionField : expansionFields) {
+      if (topLevelConfigs == null || !topLevelConfigs.containsKey(expansionField)) {
+        log.warn("Config map does not contain {}", expansionField);
+        continue;
+      }
+      p.load(new StringReader(topLevelConfigs.get(expansionField)));
```
ryannedolan, line 34: I guess this logic would repeat in any `ConfigProvider` impl. Probably better to have the SPI just return the raw properties, and then have `ConfigService` expand them this way. That way the SPI is easier to impl, and the expansion logic is in one place.

ryannedolan (approved):
Neat! Couple notes but not blocking.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sConfigProvider.java
jogrogan: Good point, though this does make throwing an exception instead trickier since there could be multiple config providers but we actually only intend to expand the properties coming from one them

PR #84, opened 2025-01-15
Add support for external engines

ryannedolan:
## Summary

This adds support for installing remote query engines, e.g. Trino, DuckDB, or Flink SQL Gateway.

 - Added Engine CRD.
 - Added `k8s.engines` metadata table.
 - Added `RemoteTableScan`, `RemoteJoin`, associated optimizer rules.

## Details

The Hoptimator JDBC Driver is able to talk to remote `Databases`, but it previously relied on Calcite's `Enumerable` engine to process queries locally. For example, joining tables in two different `Databases` would involve first fetching the rows from each table and then joining locally in the driver itself.

With `Engines`, we can outsource these operations to fast, distributed query engines like Trino. Queries are sent off to the remote engine, and the Driver simply collects the results.

We don't expect remote engines to have the same "catalog" of databases and tables that Hoptimator knows about -- Hoptimator's idea of `foo.bar` may or may not match Trino's, for example. For this reason, we fully-specify the tables used by each query, sending DDL statements ahead of each query. To do this, we leverage the same `PipelineRel` mechanism that we use for deploying a pipeline. This means that the remote query will use the same connectors and configuration as an equivalent pipeline.

## Testing

Without an `Engine` installed, a query must be processed locally via the `Enumerable` convention:

```
0: Hoptimator> explain plan for select * from ads.ad_clicks, profile.members;
PLAN  EnumerableNestedLoopJoin(condition=[true], joinType=[inner])
  JdbcToEnumerableConverter
    JdbcTableScan(table=[[ADS, AD_CLICKS]])
  JdbcToEnumerableConverter
    JdbcTableScan(table=[[PROFILE, MEMBERS]])


1 row selected (0.062 seconds)
```

The `EnumerableNestedLoopJoin` would be very slow for large datasets.

After installing an engine, we see that the query plan now involves a `RemoteJoin` instead:

```
0: Hoptimator> explain plan for select * from ads.ad_clicks, profile.members;
PLAN  RemoteToEnumerableConverter
  RemoteJoin(condition=[true], joinType=[inner])
    PipelineTableScan(table=[[ADS, AD_CLICKS]])
    PipelineTableScan(table=[[PROFILE, MEMBERS]])


1 row selected (0.041 seconds)
```

The `RemoteJoin` is able to leverage Trino or similar distributed query engines. For example, we are able to leverage our Flink session cluster via the Flink SQL gateway:

```
0: Hoptimator> select * from ads.ad_clicks, profile.members;
+------------------------------------------------------------------------------+
|                                             CAMPAIGN_URN                     |
+------------------------------------------------------------------------------+
| dad91062e84d5979ec74b5b0d0bf8e838158bed8de7a5f1b7aa0e90ffc45ea47767945bbe577 |
| f2dddb0778af2fbd2d79cd6950d44e26112413443dce63c2673c30949a99a469201388c84485 |
+------------------------------------------------------------------------------+
```

In the above results, the data is generated via Flink's `datagen` connector.


hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sEngine.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @Override
+  public DataSource dataSource() {
+    // TODO support username, password via Secrets
+    return JdbcSchema.dataSource(url, driver, null, null);
+  }
+
+  @Override
+  public SqlDialect dialect() {
+    return SqlDialect.FLINK;  // TODO fix hardcoded dialect
```
jogrogan, line 38: Can't this just return `this.dialect`?

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sEngineTable.java
```diff
@@ -0,0 +1,91 @@
...
+
+public class K8sEngineTable extends K8sTable<V1alpha1Engine, V1alpha1EngineList, K8sEngineTable.Row> {
+
+  // CHECKSTYLE:OFF
+  public static class Row {
+    public String NAME;
+    public String URL;
+    public String DIALECT;
+    public String DRIVER;
+    public String[] DATABASES;
```
jogrogan, line 31: nit: should prob be lowercase so we don't need to disable checkstyle

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/EngineRules.java
```diff
@@ -0,0 +1,164 @@
...
+  public EngineRules(Engine engine) {
+    this.engine = engine;
+  }
+
+  public void register(HoptimatorJdbcConvention inTrait, RelOptPlanner planner) {
+    SqlDialect dialect = dialect(engine);
+    String name = engine.engineName() + "-" + inTrait.database();
+    JdbcConvention outTrait = JdbcConvention.of(dialect, inTrait.expression, name);
+ 
+    System.out.println("Registering rules for " + name + " using dialect " + dialect.toString());
```
jogrogan, line 54: nit: use logger or remove (applicable to other lines in this file as well)

jogrogan (commented):
Is the planner able to distinguish when it should and shouldn't use these remote rules?

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sEngine.java
ryannedolan: yup whoops

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sEngineTable.java
ryannedolan: This is a quirk of Calcite. These fields are referenced in the generated code, and things go sideways if they aren't all caps.

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/EngineRules.java
ryannedolan: will do

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/RemoteToEnumerableConverter.java
```diff
@@ -0,0 +1,390 @@
...
+                            Expressions.block(
+                                Expressions.tryCatch(
+                                    builder.toBlock(),
+                                    Expressions.catch_(
+                                        e_,
+                                        Expressions.throw_(
+                                            Expressions.new_(
+                                                RuntimeException.class,
+                                                e_)))))))),
+                resultSet_));
```
jogrogan, line 193: 🚀

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/RemoteToEnumerableConverter.java
```diff
@@ -0,0 +1,390 @@
...
+              Expressions.call(BuiltInMethod.CREATE_ENRICHER.method,
+                  Expressions.newArrayInit(Integer.class, 1,
+                      toIndexesTableExpression(sqlString)),
+                  DataContext.ROOT));
+
+      enumerable =
+          builder0.append("enumerable",
+              Expressions.call(
+                  BuiltInMethod.RESULT_SET_ENUMERABLE_OF_PREPARED.method,
+//                  Schemas.unwrap(convention.expression(), DataSource.class),
```
jogrogan, line 218: This still needed?

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/planner/RemoteToEnumerableConverter.java
```diff
@@ -0,0 +1,390 @@
...
+                  dataSource,
+                  sql_,
+                  rowBuilderFactory_,
+                  preparedStatementConsumer_));
+    } else {
+      enumerable =
+          builder0.append("enumerable",
+              Expressions.call(
+                  BuiltInMethod.RESULT_SET_ENUMERABLE_OF.method,
+//                  Schemas.unwrap(convention.expression(), DataSource.class),
```
jogrogan, line 228: ditto

jogrogan (commented):
Super cool functionality

PR #85, opened 2025-01-15
Set up Flink session cluster and SQL gateway

jogrogan:


hoptimator-kafka/src/test/resources/kafka-ddl.id
```diff
@@ -3,24 +3,11 @@
 
 insert into kafka."existing-topic-1" select * from kafka."existing-topic-2";
 apiVersion: flink.apache.org/v1beta1
-kind: FlinkDeployment
+kind: FlinkSessionJob
```
ryannedolan, line 6: neat!

ryannedolan (approved):

PR #86, opened 2025-01-16
Fix various issues with SQL generation, flink job routing, integration-tests

jogrogan:
- Fixed issue with FLINK SQL syntax where query table: `KAFKA`.`existing-topic-2` should have just been `existing-topic-2`.
- Got deployment of Flink jobs working through Flink deployment cluster
  - Updated Flink connector options in `kafkadb.yaml` to the minimum required set needed to pass through to Flink
  - Needed to change Kafka connection mechanism. Connection on the command line using the forwarded port 9092 works via localhost to the 'external' strimzi cluster. However, this doesn't work for k8s pod to pod communication (Flink deployment pod -> Kafka cluster), pod connections need to use the 'internal' strimzi connection mechanisms `one-kafka-bootstrap.kafka.svc.cluster.local:9094`.
- Integration tests has apparently always been failing but were incorrectly showing as passing due to the `... || kill cat port-forward.pid` bit.
  - Fixed kind cluster routing that was causing these failures, also removes the ability to false report passing.
  - Added artifacts to the integration test workflow so on failure you can download the failure report since errors aren't printed to the terminal.
- hoptimator-operator was failing on startup because of two separate issues:
  - rbac was missing the "pipelines" resources
  - The hoptimator-operator K8s pod doesn't have access to .kube/config, needed its own mechanism to determine namespace information.

Testing:

Set up flink datagen connector to generate data and insert it into existing-topic-1.
Ran:
```
0: Hoptimator> create or replace materialized view kafka."existing-topic-2" as select * from kafka."existing-topic-1";
```
which setup a new Flink job to insert the data from existing-topic-1 into existing-topic-2. Was able to verify the data was inserted via CLI Kafka consumer.

Put many helpful commands in the README

Also `make integration-tests-kind` works locally as well if you setup a kind cluster locally.

### DIFF
--- a/etc/cluster.yaml
+++ b/etc/cluster.yaml
@@ -2,5 +2,17 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 # Additional worker nodes will not add any real resource capacity. However,
 # it should mean that there are more CPU reservations to hand out.
+networking:
+  ipFamily: ipv4
+  apiServerAddress: 127.0.0.1
 nodes:
 - role: control-plane
+  extraPortMappings:
+    - containerPort: 31092
+      hostPort: 9092
+      listenAddress: "127.0.0.1"
+      protocol: TCP
+    - containerPort: 31234
+      hostPort: 31234
+      listenAddress: "127.0.0.1"
+      protocol: TCP
--- a/hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sContext.java
+++ b/hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sContext.java
@@ -4,6 +4,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.Reader;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Duration;
 import java.util.Optional;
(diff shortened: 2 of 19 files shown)


ryannedolan (approved):
🔥🔥🔥

etc/cluster.yaml
```diff
@@ -2,5 +2,17 @@ kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
 # Additional worker nodes will not add any real resource capacity. However,
 # it should mean that there are more CPU reservations to hand out.
+networking:
+  ipFamily: ipv4
+  apiServerAddress: 127.0.0.1
 nodes:
 - role: control-plane
+  extraPortMappings:
```
ryannedolan, line 10: this seems important :)

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sContext.java
```diff
@@ -101,13 +103,26 @@ public static void useContext(K8sContext context) {
...
+    } else {
+      ApiClient apiClient = Config.defaultClient();
+      String filePath = System.getenv("POD_NAMESPACE_FILEPATH");
+      String namespace;
+      if (filePath == null) {
+        namespace = "default";
+      } else {
+        namespace = new String(Files.readAllBytes(Paths.get(filePath)));
+      }
+      return new K8sContext("default", namespace, apiClient);
```
ryannedolan, line 125: wow totally missed this

ryannedolan (approved):

PR #89, opened 2025-01-21
Fix various issues (see description)

jogrogan:
- Remove duplicate SqlJob definitions
  - Add configs option and status sql
  - Regenerate models
-  Fix snakeyaml dependency issue (resolving v1 and v2 causes issues)
- Fix PIpelineReconciller registration to work with HoptimatorOperatorApp (until eventual deprecation)
- Add optional namespace to ConfigProvider loading. Intended use is by an operator loading a config in another namespace.


hoptimator-util/src/main/java/com/linkedin/hoptimator/util/ConfigService.java
```diff
@@ -16,16 +17,17 @@ public final class ConfigService {
   private ConfigService() {
   }
 
+  // Null namespace will default to current namespace, may not be used by some ConfigProviders.
   // Loads top level configs and expands input fields as file-like properties
   // Ex:
   //  log.properties: |
   //    level=INFO
-  public static Properties config(String... expansionFields) {
+  public static Properties config(@Nullable String namespace, String... expansionFields) {
```
ryannedolan, line 25: Just thinking -- would it be possible to intuit the `expansionFields` via some convention? e.g. if the key ends in `.properties` automatically expand it?

ryannedolan (approved):

hoptimator-util/src/main/java/com/linkedin/hoptimator/util/ConfigService.java
jogrogan: Yea it would, we can talk about making this process better.
Right now it is loading the configs pretty much anytime it needs to without any caching. The point of the expansion fields was that if we are doing it this way we won't need to resolve nested properties each time unless they are actually needed. i.e. when xinfra.properties is needed, don't need to expand venice.properties.

There's of course pros/cons with a caching approach as well. it's nice right now because we can add configs on the fly without needing to redeploy an operator

PR #91, opened 2025-01-22
Fix compound identifier bug

ryannedolan:
## Summary

- Added `CREATE DATABASE` statements to generated SQL
- Added schema name to `CREATE TABLE` and `INSERT INTO` SQL
- Dropped extraneous logic unflattening logic

## Details

The old planner generated `CREATE IF NOT EXISTS DATABASE` statements along with `CREATE IF NOT EXISTS TABLE` statements, allowing tables to be referenced as `<schema>.<table>`. I had dropped this logic in the new version of the planner, but it ends up being useful. Without fully-qualified table names, it becomes difficult to distinguish between tables, fields, and `t1.foo` field aliases in generated SQL. In particular, joins across tables with the same field name where incorrectly being rewritten like `foo = foo`, instead of `t1.foo = t2.foo`.

## Testing

Integration tests pass with and without the Flink engine installed.

Manually tested a query with the fink engine installed:

```
0: Hoptimator> select * from ads.audience;
INFO Open session to http://localhost:8083 with connection version: V2.
INFO DDL: CREATE DATABASE IF NOT EXISTS `ADS` WITH ()
INFO DDL: CREATE TABLE IF NOT EXISTS `ADS`.`PAGE_VIEWS` (`PAGE_URN` VARCHAR, `MEMBER_URN` VARCHAR) WITH ('connector'='datagen', 'number-of-rows'='10')
INFO DDL: CREATE DATABASE IF NOT EXISTS `PROFILE` WITH ()
INFO DDL: CREATE TABLE IF NOT EXISTS `PROFILE`.`MEMBERS` (`FIRST_NAME` VARCHAR, `LAST_NAME` VARCHAR, `MEMBER_URN` VARCHAR, `COMPANY_URN` VARCHAR) WITH ('connector'='datagen', 'number-of-rows'='10')
INFO SQL: SELECT * FROM (SELECT `MEMBER_URN`
         FROM `ADS`.`PAGE_VIEWS`) AS `t`
     INNER JOIN (SELECT `FIRST_NAME`, `LAST_NAME`, `MEMBER_URN`
         FROM `PROFILE`.`MEMBERS`) AS `t0` ON `t`.`MEMBER_URN` = `t0`.`MEMBER_URN`
+------------+-----------+
| FIRST_NAME | LAST_NAME |
+------------+-----------+
No rows selected (52.114 seconds)
```

N.B. the query references view `ads.audience`, which is properly expanded as a join in the generated SQL. The join runs now.

### DIFF
--- a/hoptimator-kafka/src/test/resources/kafka-ddl.id
+++ b/hoptimator-kafka/src/test/resources/kafka-ddl.id
@@ -12,9 +12,11 @@ spec:
   job:
     entryClass: com.linkedin.hoptimator.flink.runner.FlinkRunner
     args:
-    - CREATE TABLE IF NOT EXISTS `existing-topic-2` (`KEY` VARCHAR, `VALUE` BINARY) WITH ('connector'='kafka', 'properties.bootstrap.servers'='one-kafka-bootstrap.kafka.svc.cluster.local:9094', 'scan.startup.mode'='earliest-offset', 'topic'='existing-topic-2', 'value.format'='json')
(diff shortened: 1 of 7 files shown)


jogrogan (approved):
Nice!

PR #95, opened 2025-01-24
Resolve complex array types

jogrogan:
Made various changes:

- Added nullability component into our AvroConverter to make the `RelDataType` we generate work with the JDBC schema conversion happening in Calcite.
- Added flink configs to the status of the SqlJob to be able to determine if a resource diverged based on its config and needs to be redeployed
- Added ability into ConfigService to only load expansion fields. This is needed by Flink to only get the properties under `flink.config`
  - Added ability to load these properties into the Job template
- Added the ability to flatten complex arrays into the sum of its parts and the ability to unflatten it back into a complex array
- Fix various checkstyle complaints


ryannedolan (approved):
nice

PR #97, opened 2025-01-29
Fix default watch namespacing

jogrogan:
- Added consistency for namespace usage between `Operator` and `K8sContext`
- Changed default namespace to "hoptimator-operator" when no kube config is defined. May want to introduce a way to override this default
- nit: put KEY columns first for Venice


hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sContext.java
```diff
@@ -105,27 +125,29 @@ public static void useContext(K8sContext context) {
     currentContext = context;
   }
 
-  static K8sContext defaultContext() throws IOException {
+
+  // If $HOME/.kube/config is defined, use that config file.
+  // If POD_NAMESPACE_FILEPATH is defined, use that config file.
+  // Use Config.defaultClient() and defaultNamespace if no config file is found.
+  static K8sContext defaultContext(String defaultNamespace) throws IOException {
```
ryannedolan, line 132: Hmm, I think the default namespace is fine to keep "default". I don't see much utility in making this configurable at runtime. It should be the _current_ namespace, or "default" failing that. I don't think we need any further flexibility here.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sContext.java
```diff
@@ -56,9 +63,18 @@ public SharedInformerFactory informerFactory() {
     return informerFactory;
   }
 
+  public String watchNamespace() {
+    return watchNamespace;
+  }
+
+  public void watchNamespace(String watchNamespace) {
+    this.watchNamespace = watchNamespace;
```
ryannedolan, line 71: I think it's best if `currentContext()` is a global variable / singleton, without any setters. It's unpredictable otherwise. e.g. you could invoke `watchNamespace("foo")` and have no idea if your change is before or after the reconcilers start listening.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sContext.java
```diff
@@ -56,9 +63,18 @@ public SharedInformerFactory informerFactory() {
...
+  }
+
   public <T extends KubernetesObject, U extends KubernetesListObject> void registerInformer(
       K8sApiEndpoint<T, U> endpoint, Duration resyncPeriod) {
-    informerFactory.sharedIndexInformerFor(generic(endpoint), endpoint.elementType(), resyncPeriod.toMillis());
+    LOG.info("Registering resource={} on namespace={}", endpoint.kind(), namespace);
+    informerFactory.sharedIndexInformerFor(generic(endpoint), endpoint.elementType(), resyncPeriod.toMillis(), watchNamespace);
```
ryannedolan, line 77: Can we just add `watchNamespace` as a parameter to `registerInformer`? This method is only invoke from the operator, which now knows what namespace to watch via the command-line option.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/HoptimatorOperatorApp.java
```diff
@@ -59,9 +61,15 @@ public static void main(String[] args) throws Exception {
 
     Options options = new Options();
 
-    Option namespace = new Option("n", "namespace", true, "specified namespace");
-    namespace.setRequired(false);
-    options.addOption(namespace);
+    Option defaultNamespace = new Option("n", "namespace", true,
```
ryannedolan, line 64: This doesn't make sense to me.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/HoptimatorOperatorApp.java
```diff
@@ -59,9 +61,15 @@ public static void main(String[] args) throws Exception {
...
+        "default namespace to configure K8s API");
+    defaultNamespace.setRequired(false);
+    options.addOption(defaultNamespace);
+
+    Option watchNamespace = new Option("w", "watch", true,
```
ryannedolan, line 69: This does make sense. Let's keep the watch argument, and just pass this through to the `informers`.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/PipelineOperatorApp.java
```diff
@@ -18,12 +25,50 @@
...
+  public PipelineOperatorApp(String defaultNamespace, String watchNamespace) {
+    this.defaultNamespace = defaultNamespace;
+    this.watchNamespace = watchNamespace;
+  }
+
   public static void main(String[] args) throws Exception {
-    new PipelineOperatorApp().run();
+    Options options = new Options();
+
+    Option defaultNamespace = new Option("n", "namespace", true,
```
ryannedolan, line 39: drop

ryannedolan (changes requested):
I don't think making the "namespace" configurable makes sense. There are only two possible situations: the operator is running inside K8s (in which case it will know what namespace to use), or it is running outside. If outside, we already have `kube/config`, which lets you configure any namespace. I don't think overriding that logic makes sense.

That said, adding a `watch` argument makes sense. Operators typically do that. I think we should just thread that through to `registerInformer` rather than muck with `K8sContext`, since that is meant to be a read-only singleton.

ryannedolan (approved):
Love it, thx

PR #98, opened 2025-02-03
Add support for complex union types and maps

jogrogan:
Requires calcite version bump and kubernetes version bump. But this also solves all of the snakeyaml compat issues we were running into


ryannedolan (approved):
Probably worth a unit test. lgtm.

PR #99, opened 2025-02-05
Add per-connection driver properties

ryannedolan:
## Summary

- Added `connectionProperties` parameter to several API methods, including `DeployerProvider` and `ConfigProvider`.
- `K8sContext` now constructs an `ApiClient` from the `connectionProperties`, or failing that, from `kube/config`.
- Dropped `defaultContext()` and related singleton methods, ensuring the driver is re-entrant and can handle simultaneous connections with different configurations.

## Details

Previously, we treated `K8sContext` as a lazy-loaded singleton, which was shared across connections. This was preventing us from productionizing the driver in a multi-tenant service, where different clients may want to use distinct namespaces, usernames, etc. To solve this problem, we now construct `K8sContext` dynamically from connection-level properties. This required some minor API changes, in order to thread the connection properties through to the various plugins. In particular, `DeployerProvider` and `ConfigProvider` now have access to the connection properties. `K8sDeployer` etc use this to construct `ApiClient`s etc.

## Testing

Verified the driver loads from the CLI, unit tests pass, and integration tests pass.


hoptimator-jdbc/build.gradle
```diff
@@ -54,6 +54,16 @@ java {
   targetCompatibility = JavaVersion.VERSION_11
 }
 
+// Work-around for gradle text fixtures bug, which sponteously started publishing
```
jogrogan, line 57:
```suggestion
// Work-around for gradle test fixtures bug, which spontaneously started publishing
```

jogrogan (approved):
LGTM, we should probably at some point start a wiki to document some of these properties we have defined

PR #102, opened 2025-02-09
Simplify config loading

ryannedolan:
## Summary

- Do not load JDBC driver configuration from `ConfigService`. Instead, rely on connection properties.
- Add `catalogs` connection property to limit loaded catalogs.
- Properly load connection properties from JDBC URL or via `getConnection()` argument.

## Details

Previously, `ConfigService` was used within various plugins (e.g. JDBC drivers) to load dynamic configuration. Some of these cases are now obviated by #99's connection properties, which thread through to each plugin, _except_ JDBC drivers. For JDBC drivers, we prefer to use their respective URL, when possible, since relying on `ConfigService` may involve dynamically fetching configurations (e.g. ConfigMaps) when loading each database.

In addition, we will want to support a `CREATE DATABASE` eventually, which won't be able to provide the out-of-band configuration that `ConfigService` needs. For example, we can have `create database foo.bar location 'jdbc:foo://cluster=bar;fabric=prod'`, which fully specifies the external connection. We don't need to additionally supply a configmap out-of-band. 

## Testing

Verified manually with client.

### DIFF
--- a/hoptimator-kafka/src/main/java/com/linkedin/hoptimator/kafka/KafkaDriver.java
+++ b/hoptimator-kafka/src/main/java/com/linkedin/hoptimator/kafka/KafkaDriver.java
@@ -11,8 +11,6 @@ import org.apache.calcite.jdbc.CalciteConnection;
 import org.apache.calcite.jdbc.Driver;
 import org.apache.calcite.schema.SchemaPlus;
 
-import com.linkedin.hoptimator.util.ConfigService;
-
 
 /** JDBC driver for Kafka topics. */
 public class KafkaDriver extends Driver {
@@ -36,9 +34,9 @@ public class KafkaDriver extends Driver {
     if (!url.startsWith(getConnectStringPrefix())) {
       return null;
     }
-    // Connection string properties are given precedence over config properties
-    Properties properties = ConfigService.config(null);
+    Properties properties = new Properties();
     properties.putAll(ConnectStringParser.parse(url.substring(getConnectStringPrefix().length())));
+    properties.putAll(props); // in case the driver is loaded via getConnection()
     try {
       Connection connection = super.connect(url, props);
       if (connection == null) {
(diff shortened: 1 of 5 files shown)


hoptimator-kafka/src/main/java/com/linkedin/hoptimator/kafka/KafkaDriver.java
```diff
@@ -36,9 +34,9 @@ public Connection connect(String url, Properties props) throws SQLException {
     if (!url.startsWith(getConnectStringPrefix())) {
       return null;
     }
-    // Connection string properties are given precedence over config properties
-    Properties properties = ConfigService.config(null);
+    Properties properties = new Properties();
     properties.putAll(ConnectStringParser.parse(url.substring(getConnectStringPrefix().length())));
+    properties.putAll(props); // in case the driver is loaded via getConnection()
```
jogrogan, line 39: Should these lines be reversed? Trying to think if there are config collisions, which should take precedence. I'd think the connection properties should (same as you have for HoptimatorDriver)

hoptimator-venice/src/main/java/com/linkedin/hoptimator/venice/VeniceDriver.java
```diff
@@ -40,8 +38,9 @@ public Connection connect(String url, Properties props) throws SQLException {
       return null;
     }
     // Connection string properties are given precedence over config properties
-    Properties properties = ConfigService.config(null, CONFIG_NAME);
+    Properties properties = new Properties();
     properties.putAll(ConnectStringParser.parse(url.substring(getConnectStringPrefix().length())));
+    properties.putAll(props); // in case the driver is loaded via getConnection()
```
jogrogan, line 43: ditto

jogrogan (approved):

hoptimator-kafka/src/main/java/com/linkedin/hoptimator/kafka/KafkaDriver.java
ryannedolan: The JDBC spec says the precedence order is undefined and driver specific! So whatever makes sense to us I guess. Agree that the URL should take precedence, since that is more likely to be what end users are actually able to manipulate. Will fix, thx.

PR #105, opened 2025-02-13
Fix Zeppelin k8s connection

jogrogan:
Likely will need to make similar new yaml files to get Kafka & Venice working through Zeppelin, will add separately if needed


ryannedolan (approved):

ehoner (approved):
🚀

PR #106, opened 2025-02-13
Remove incorrect operator namespace

jogrogan:
Watch namspace != the namespace we want to connect to. Removing this will default the operator pod to use environment variable: `POD_NAMESPACE_FILEPATH`


ryannedolan (approved):
Whoops!

PR #110, opened 2025-02-14
Apply some Zeppelin fixes for kafka

jogrogan:
- Connection properties were not being loaded all the way through properly
- Kafka bootstrapUrl needed to be changed to work within docker
- Noticed some rbac discrepancies
- Needed to add "key.format" for flink kafka to work
- Left in a TODO around kafka broker advertised ports. Tried adding a new broker and a new listener, haven't gotten it to work yet.

### DIFF
--- a/deploy/dev/kafka.yaml
+++ b/deploy/dev/kafka.yaml
@@ -45,6 +45,7 @@ spec:
             nodePort: 31092
           brokers:
           - broker: 0
+#            advertisedHost: host.docker.internal # swap these lines to enable Zeppelin, TODO: figure out a way around this
             advertisedHost: 127.0.0.1
             nodePort: 31234
     config:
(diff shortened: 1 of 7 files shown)


deploy/dev/kafka.yaml
```diff
@@ -45,6 +45,7 @@ spec:
             nodePort: 31092
           brokers:
           - broker: 0
+#            advertisedHost: host.docker.internal # swap these lines to enable Zeppelin, TODO: figure out a way around this
```
ryannedolan, line 48: We could theoretically have multiple listeners here, one for the `host.docker.internal` use-case, and one for the `localhost` use-case. For example, maybe `external` and `docker` could be separate "listeners" on different ports.

But the point is kinda moot of we just move Zeppelin inside K8s. All this stuff gets easier.

ryannedolan (approved):

deploy/dev/kafka.yaml
jogrogan: Yea I tried this and it wouldn't work... didn't want to kill too much time on it

PR #111, opened 2025-02-20
Add ability to drop views

jogrogan:
Add ability to drop views and materialized views

Tested a few cases:
```
drop materialized view ADS."ad_clicks$test1"; // Works if exists, errors if not
drop materialized view IF EXISTS ADS."ad_clicks$test1"; // Always works
drop view ads.MY_VIEW; // Works if exists, errors if not
drop view IF EXISTS ads.MY_VIEW; // Always works
```

Added some integration tests


ryannedolan (approved):
Nice! Cool to see the integration tests clean up after themselves :)

Please add a note that this code was copy-pasted from upstream.

PR #112, opened 2025-02-24
Fix connection pass through preventing functions from working

jogrogan:
Fixes a few issues:
- Make demodb consistent in using key/value pair for jdbc connection
- Introduce HoptimatorViewTableMacro, this is strictly to work around the static connection object Calcite uses for materialized views (don't really get why they do this): https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/schema/impl/MaterializedViewTable.java#L58
- Add property support for quidem testing
- Update K8sDatabaseTable/K8sViewTable to also pass through connection properties

### DIFF
--- a/hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sDatabaseTable.java
+++ b/hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sDatabaseTable.java
@@ -3,6 +3,7 @@ package com.linkedin.hoptimator.k8s;
 import java.util.Locale;
 import java.util.Optional;
 import java.util.Properties;
+import java.util.StringJoiner;
 import javax.sql.DataSource;
 
 import org.apache.calcite.adapter.jdbc.JdbcSchema;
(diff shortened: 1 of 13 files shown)


hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sDatabaseTable.java
```diff
@@ -82,9 +83,28 @@ private static String schemaName(Row row) {
...
+        String value = connectionProperties.getProperty(key);
+        joiner.add(key + "=" + value);
+      }
+    }
+    String joinedUrl = row.URL;
+    // Handles case where there are no properties already in the URL
+    if (row.URL.endsWith("//")) {
+      joinedUrl = joinedUrl + joiner;
+    } else {
+      joinedUrl = joinedUrl + ";" + joiner;
```
ryannedolan, line 105: The `k1=v1;k2=v2` format seems to be a common convention, but not all drivers support it. We may run into drivers that choke on it. Not sure what to do in that case...

ryannedolan (approved):
Wow! This bug had me totally confused. Way to go.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sDatabaseTable.java
jogrogan: All the drivers we have for now do support it, so maybe that's good enough for now? Considering Properties objects are always key/value pairs I'd imagine we'd run into more problems if we just had single values

PR #113, opened 2025-02-24
Use macro override method

jogrogan:
Change HoptimatorViewTableMacro to use override method preventing any accidental calls to super.


ehoner (approved):

ehoner (approved):
LGTM

PR #114, opened 2025-02-25
Add ownership references

ryannedolan:
## Summary

- Add ownership references to pipeline elements.
- Defer specifying pipeline elements until they are deployed (instead of during planning).
- Simplify `Connector`, `Deployer`, `Validator`.
- Only support deploying objects that extend `Deployable`.
- Introduce `View` deployable.

## Details

`K8sMaterializedViewDeployer` now deploys the `View` and `Pipeline` objects, rather than having separate hooks deploying each. This enables us to create the `Pipeline` with an owner reference to the `View` object. It then goes on to deploy all the elements of the pipeline, creating/updating elements as needed. Each element is created/updated to include an owner reference to the `Pipeline` object.

Ownership thus is `View` --> `Pipeline` --> elements.

The deployer also tags deployed resources with a `pipeline` label, though this is not particularly useful.

Several small API changes were made to make this slightly easier.

N.B. ownership references cannot span namespaces. Creating an object with a non-existing owner causes the object to be immediately garbage-collected. This logic also applies if we try to create child objects in a different namespace, since the owner won't be found. To prevent this situation, the deployer enforces that pipeline elements all belong to the same namespace as the pipeline. In order to span namespaces, an operator _must_ be used in between. For example, I've configured the K8s operator to watch the `default` namespace, while deploying Kafka itself in the `kafka` namespace. This means pipelines in `default` can effect Kafka in `kafka`, but only because there is an operator in between.

## Testing

```
0: Hoptimator> create or replace materialized view kafka."blam" as select * from kafka."existing-topic-1";
No rows affected (2.154 seconds)
```

Then:

```
% kubectl tree view kafka-blam
NAMESPACE  NAME                                            READY  REASON            AGE
default    View/kafka-blam                                 -                        5m5s
default    └─Pipeline/kafka-blam                           -                        5m5s
default      ├─FlinkSessionJob/kafka-database-blam         -                        5m5s
default      ├─KafkaTopic/kafka-database-blam              True                     5m5s
default      └─KafkaTopic/kafka-database-existing-topic-1  False  ResourceConflict  5m5s
```

Also verified that `drop view` cleans up the entire tree.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,8 +103,8 @@ undeploy-dev-environment: undeploy-venice undeploy-kafka undeploy-flink undeploy
 # Integration test setup intended to be run locally
 integration-tests: deploy-dev-environment
 	kubectl wait kafka.kafka.strimzi.io/one --for=condition=Ready --timeout=10m -n kafka
-	kubectl wait kafkatopic.kafka.strimzi.io/existing-topic-1 --for=condition=Ready --timeout=10m -n kafka
-	kubectl wait kafkatopic.kafka.strimzi.io/existing-topic-2 --for=condition=Ready --timeout=10m -n kafka
+	kubectl wait kafkatopic.kafka.strimzi.io/existing-topic-1 --for=condition=Ready --timeout=10m
+	kubectl wait kafkatopic.kafka.strimzi.io/existing-topic-2 --for=condition=Ready --timeout=10m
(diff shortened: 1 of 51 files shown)


jogrogan (approved):
LGTM! 👏 👏

PR #115, opened 2025-02-26
Register materialized views with the planner

ryannedolan:
## Summary

- Enable catalogs to register materializations with the planner.
- Delay validation of view SQL until after all views have loaded.
- Minor API changes to carry around a `Connection`.

## Details

Calcite supports substituting all or parts of a query with a view, which may accelerate queries if the view is materialized. We want Hoptimator to leverage this mechanism when planning a pipeline, which would enable, for example, a pipeline to read from a snapshot rather than a source-of-truth event stream. In that scenario, the snapshot must be registered as a materialization of the event stream.

To that end, these changes enable catalogs to register arbitrary materializations. The K8s catalog naively registers all materialized views this way. We can also easily envision a catalog that registers Iceberg tables corresponding to every Kafka topic, etc.

I haven't been able to verify that the planner actually considers using the materializations, but at least it doesn't complain.

## Testing

Existing tests pass. Leaving validation of materializations to future work.


jogrogan (commented):
Cool stuff! looks like some tests are failing but otherwise LGTM

jogrogan (approved):

PR #116, opened 2025-02-26
Support hints via jdbc connection properties

jogrogan:
The main change in this PR is to add a hints KV field to the jdbc connection properties. This allows users to side-load hints. Ideally, users would be able to pass hints through the SQL itself [see here](https://calcite.apache.org/docs/reference.html#sql-hints) but Calcite does not support hints on materialized views.

I did explore a separate path of being able to list all hints through SQL on the SELECT statement however I don't think this solution is any better: https://github.com/linkedin/Hoptimator/compare/main...jogrogan/hints

This change is kept pretty simple so that we can change it in the future should we want to go with the above solution or should we want to have separate hints for sources/sinks/jobs.

List of changes:
- Add hints to Sink & JobDeployer, still working on adding it to the connector deployer (will come in a separate PR)
- Fixed bug with PipelineReconciler throwing exception on deletion
- Expanded README
- Fixed checkstyle complaints


ryannedolan (approved):
Nice! This makes sense to me, and jibes well with our internal tooling. Hopefully I didn't step too hard on this PR with my last one :)

PR #117, opened 2025-02-26
Support connector hints

jogrogan:
Two changes:
- Adds hints to source objects to show up in the connector. It's a very real use case where a user may want to pass through hint information maybe as to the consumer group or something else configurable in flink.
- Passes through a "pipelineName" that will show up in the source/sink. This will be necessary for template parameters that require pipeline-uniqueness.

### DIFF
--- a/hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorDdlExecutor.java
+++ b/hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorDdlExecutor.java
@@ -189,9 +189,9 @@ public final class HoptimatorDdlExecutor extends ServerDdlExecutor {
       // is "foo-bar" and the sink is just FOO.
       String[] viewParts = viewName.split("\\$", 2);
       String sinkName = viewParts[0];
-      String partialViewName = "";
+      String pipelineName = database + "-" + sinkName;
       if (viewParts.length > 1) {
-        partialViewName = viewParts[1];
+        pipelineName = pipelineName + "-" + viewParts[1];
       }
       List<String> sinkPath = new ArrayList<>();
       sinkPath.addAll(schemaPath);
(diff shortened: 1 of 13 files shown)


ryannedolan:
re `viewName`, `K8sMaterializedViewDeployer` sets `{{name}}` for pipeline uniqueness purposes. Do we need more than that?

jogrogan:
> re `viewName`, `K8sMaterializedViewDeployer` sets `{{name}}` for pipeline uniqueness purposes. Do we need more than that?

Name is the name of the sink, which doesn't include the partial view information:
Ex:
```
create or replace materialized view kafka."existing-topic-1$test" as select * from kafka."existing-topic-2";
```
yields {{name}} == "kafka-database-existing-topic-2"

If you have many partial views on the same topic, it won't be unique

ryannedolan:
> yields {{name}} == "kafka-database-existing-topic-2"

It _should_ yield `kakfa-database-existingtopic2-test`. Maybe we broke that? I bet I broke it when I changed `K8sMaterializedViewDeployer`.

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorDdlExecutor.java
```diff
@@ -187,7 +187,12 @@ public void execute(SqlCreateMaterializedView create, CalcitePrepare.Context con
 
       // Support "partial views", i.e. CREATE VIEW FOO$BAR, where the view name
       // is "foo-bar" and the sink is just FOO.
-      String sinkName = viewName.split("\\$", 2)[0];
+      String[] viewParts = viewName.split("\\$", 2);
+      String sinkName = viewParts[0];
+      String partialViewName = "";
```
ryannedolan, line 192: We should use `null` instead of empty string.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sJobDeployer.java
```diff
@@ -41,6 +41,7 @@ public List<String> specify() throws SQLException {
         .with("database", job.sink().database())
         .with("schema", job.sink().schema())
         .with("table", job.sink().table())
+        .with("partialViewName", job.sink().partialViewName())
```
ryannedolan, line 44: Don't recall how `.with(x, null)` works. We may want to add `.withDefault(x, y, "")` or something like that.

ryannedolan (commented):
I think `partialViewName` is a bit leaky, since the planner doesn't otherwise know about partial views. Could we get away with `pipelineName` here?

For the intended use-cases (e.g. consumer groups, client IDs, etc), can they just use the full pipeline name, e.g. `kafka-database-existingtopic1-foo`?

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorDdlExecutor.java
jogrogan: I liked the idea of pipelineName, way cleaner and now don't have to worry about nulls

ryannedolan (approved):
Sorry for sending you around in circles. Make sense, thanks.

PR #118, opened 2025-02-27
Add missing rules for LISTAGG & UNNEST

jogrogan:
SQL functions were still not fully working with #112. The tests were only testing views, creation of materialized views, when walking the implementor path, uncovered a few issues.

- Added `LogicAggregate` rule to support LISTAGG
- Added `Uncollect` rule to support UNNEST

There are still feature gaps present
```
create or replace view ADS."unnested" AS SELECT * FROM UNNEST(ARRAY(SELECT "FIRST_NAME" FROM profile.members)) AS name;
```
This still fails expected rules for `Collect` and `LogicalValues`. Hit a roadblock with `Collect` which is missing `RelToSqlConverter` conversion so gave up on that for now.


ehoner (approved):

ryannedolan (approved):
Nice work!

PR #125, opened 2025-03-24
Default to ANSI dialect

jogrogan:
I'm not convinced [MysqlSqlDialect](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/dialect/MysqlSqlDialect.java) is correct to use over [AnsiSqlDialect](https://github.com/apache/calcite/blob/main/core/src/main/java/org/apache/calcite/sql/dialect/AnsiSqlDialect.java).

Both seem to have the same quoting mechanisms set up, MySQL seems to have slight differences to types & casting (which is confirmed won't work in Flink), null handling (which shouldn't matter to us), and contains some additional functions (which are also unconfirmed will work in Flink).

### DIFF
--- a/hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sEngineTable.java
+++ b/hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sEngineTable.java
@@ -72,7 +72,7 @@ public class K8sEngineTable extends K8sTable<V1alpha1Engine, V1alpha1EngineList,
     if (row.DIALECT == null) {
       return SqlDialect.ANSI;
     }
-    return SqlDialect.valueOf(row.DIALECT);
+    return SqlDialect.valueOf(row.DIALECT.toUpperCase());
   }
 
   @Override
--- a/hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sEngineTable.java
+++ b/hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sEngineTable.java
@@ -72,12 +72,7 @@ public class K8sEngineTable extends K8sTable<V1alpha1Engine, V1alpha1EngineList,
     if (row.DIALECT == null) {
       return SqlDialect.ANSI;
     }
-    switch (row.DIALECT) {
-    case "Flink":
-      return SqlDialect.FLINK;
-    default:
-      return SqlDialect.valueOf(row.DIALECT);
-    }
+    return SqlDialect.valueOf(row.DIALECT);
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sJobDeployer.java
```diff
@@ -42,7 +42,6 @@ public List<String> specify() throws SQLException {
         .with("schema", job.sink().schema())
         .with("table", job.sink().table())
         .with("sql", sql.apply(SqlDialect.ANSI))
-        .with("flinksql", sql.apply(SqlDialect.FLINK))
```
ryannedolan, line 45: I think we still need `{{flinksql}}`, and we should use that in our flink templates. But maybe it just maps to the same thing as `{{sql}}` for now.

Keep in mind Flink is unique wrt it's DDL. We need to know we are generating Flink DDL, even if the queries are just ANSI. If/when we want to target another engine (e.g. spark) we will need something like `{{sparksql}}` that uses different DDL, even if the dialect for queries is still just ANSI.

ryannedolan: OTOH, we have been talking about dropping DDL generation, so maybe it doesn't matter longer term what we do here.

jogrogan: Fair point, I added back flinksql but just have `SqlDialect.FLINK` & `SqlDialect.ANSI` using the ANSI dialect

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sEngineTable.java
```diff
@@ -72,12 +72,7 @@ private static SqlDialect dialect(Row row) {
     if (row.DIALECT == null) {
       return SqlDialect.ANSI;
     }
-    switch (row.DIALECT) {
-    case "Flink":
-      return SqlDialect.FLINK;
```
ryannedolan, line 77: Think we still need this. If an engine has  `dialect: Flink` that needs to map to FLINK enum here.

ryannedolan: Note there are two concepts of dialect at play. Calcite has a sql dialect but we layer on our own. Our `SqlDialect.FLINK` should basically be Calcite's ANSI plus whatever Flink-specific DDL we need to generate.

ryannedolan (approved):

PR #127, opened 2025-04-07
Support connector hint passthrough

jogrogan:
Supports automatic connector hint passthrough without the need for templates.

Format is expected as `<engine-connector-name>.<source|sink>.<connectorConfigName>`

See README for more information

Also fixed a bug when applying sqljob yaml for configs where it would complain about ints or booleans because it wasn't surrounded in single quotes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,13 +192,13 @@ These fields can then be added to templates as `{{kafka.partitions}}` or `{{flin
 
 2. Connector hints allow the user to pass configurations directly through to an Engine (e.g. Flink).
 
-Connector hints must be formatted as follows `<schema>.<source|sink>.<configName>`
+Connector hints must be formatted as follows `<engine-connector-name>.<source|sink>.<configName>`
 
 For example, to set a Kafka group id and startup mode to be used by Flink, you could add the following hints to the connection:
 ```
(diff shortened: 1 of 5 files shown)


ryannedolan:
> `<schema>.<source|sink>.<connectorConfigName>`

Connector configs depend on the connector, which is hard to predict and impossible to control based on schema. For example, say `KAFKA` schema is implemented by _both_ `kafka` and `iceberg` connectors, or by `kafka` in streaming mode and `iceberg` in batch mode. The user specifying `KAFKA.sink.foo` could not be sure `foo` will get passed to a Kafka connector -- only that it will get passed to whatever is configured by the `KAFKA` table/job template(s). In particular, their hint could easily get passed to an `iceberg` connector instead.

This creates a problem, since Flink connectors will _generally_ throw an error if they see a configuration they don't understand. The `kafka` connector's `property.x.y` option is an exception to this rule -- the Kafka connector just passes these configs through to the Kafka client. So this works fine for `KAFKA.sink.property.client.id`, for example, even though there is no matching connector config. But that same config would break if it were passed to a more typical connector.

An alternative might be to use the _connector name_ rather than the schema name as a prefix. That ensures we only pass `iceberg.source.foo.bar` to an `iceberg` source connector, for example. This also feels more "hinty" to me, cuz hints will just get ignored if they aren't relevant.

Does that work?

ryannedolan (approved):
neat!

PR #128, opened 2025-04-14
Add retriable/transient exceptions

ryannedolan:
## Summary

- Throw `SQLNonTransientException` (or derivative) for validation errors, etc.
- Throw `SQLTransientConnectionException` for IO errors encountered when loading the catalog, etc.

## Details

Calcite frequently uses unchecked exceptions, and Hoptimator follows that pattern where necessary. Unfortunately, this means client code often receives opaque `RuntimeException`s. Where possible, we want to instead throw `SQL[Non]TransientException`, s.t. client code can decide whether to retry or surface an error to users.

## Testing

None. Client code will be tested externally.

N.B. we don't expect 100% coverage wrt this transient-vs-non-transient distinction at this point. As future `RuntimeException`s are encountered, we may need to revisit this problem.


jogrogan (approved):
Nice!

PR #131, opened 2025-05-05
Refactor status prediction logic in PipelineReconciler to a common package

nguyent8:
## Problem & Solution Overview

We would like to expose statues of pipeline and its elements as new K8s metadata status tables.
This PR is to refactor the common status estimating logic to be reused in another PR - see:
https://github.com/linkedin/Hoptimator/tree/trung/status-table

## Testing Done
- K8sPipelineElementStatusEstimatorTest
- Created a new pipeline in hoptimator and verified logs in hoptimator-operator-deployment pod


ryannedolan (approved):
neat!

PR #135, opened 2025-05-07
Create pipeline element status tables

nguyent8:
## Problem & Solution Overview
Currently users can only see statuses of pipelines through the "k8s".pipelines metadata table but there are cases, such as debugging, where they would like to see more details of the elements deployed for the pipelines. As of now, they need do multiple Kubernetes queries directly to get those information.

This PR is to provide a more convenient way to get status of pipelines and their direct elements by:
- adding new fields (ready, failed) to the existing new fields "k8s".pipelines
- adding new tables:
  - "k8s".pipeline_elements table: provides statuses of the elements specified in the spec field of the pipeline K8s objects
  - "k8s".pipeline_element_map : provides n:n mapping between pipeline elements and their pipelines

In the future, the above tables can be expanded to provide statuses of indirect elements, i.e. the ones deployed based on spec fields of other elements.

## Testing Done

- Integration test: com.linkedin.hoptimator.k8s.TestSqlScripts.k8sMetadataTables 

- Manual verification with the hoptimator cli:

  - deploy dev environment

  - create a couple of pipelines

  - verify statuses of pipelines and their elements by querying  "k8s".pipelines and the new tables "k8s".pipeline_elements and "k8s".pipeline_element_map, e.g.:
```
select pl.name as pipeline_name, pl.ready as pipeline_ready, pl.status as pipeline_status, pe.element_name, pe.ready as element_ready, pe.failed as element_failed, pe.status as element_status from "k8s".pipelines pl inner join (select t2.element_name, t1.ready, t1.failed, t1.status, t2.pipeline_name from "k8s".pipeline_elements t1 inner join "k8s".pipeline_element_map t2 on t1.name = t2.element_name) pe on pl.name = pe.pipeline_name;

```


hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElementMapTable.java
```diff
@@ -0,0 +1,41 @@
+package com.linkedin.hoptimator.k8s;
+
+import com.linkedin.hoptimator.util.RemoteTable;
+import org.apache.calcite.schema.Schema;
+
+
+/** Provides n:n mapping between pipeline elements and their pipelines. */
+public class K8sPipelineElementMapTable
```
jogrogan, line 8: So elements of a pipeline can definitely belong to multiple pipelines. Example: Single kafka topic being written to by two separate flink jobs.

This model would still work and support this, right?

nguyent8: yeah the backing data is provided by K8sPipelineElementMapApi which is backed by K8sPipelineElementApi

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElement.java
```diff
@@ -0,0 +1,38 @@
...
+  private final K8sPipelineElementStatus status;
+  private final Set<V1alpha1Pipeline> pipelines = new HashSet<>();
+
+  public K8sPipelineElement(V1alpha1Pipeline pipeline, K8sPipelineElementStatus status) {
+    this.name = status.getName();
+    this.status = status;
+    this.pipelines.add(pipeline);
+  }
+
+  public String getName() {
```
ryannedolan, line 23: Most of our code is more "scala" style than "bean" style. I'd call this `name()` rather than `getName()`. Mox nix.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElementApi.java
```diff
@@ -0,0 +1,54 @@
...
+    this.context = context;
+  }
+
+  private Collection<K8sPipelineElement> discoverAllElements(K8sContext context) {
+    final K8sApi<V1alpha1Pipeline, V1alpha1PipelineList> pipelineApi = new K8sApi<>(context, K8sApiEndpoints.PIPELINES);
+    Collection<V1alpha1Pipeline> pipelines;
+    try {
+      pipelines = pipelineApi.list();
+    } catch (SQLException e) {
+      throw new RuntimeException(e);
```
ryannedolan, line 29: Can we add a checked exception to this method and just let it escape to the caller? The API expects `SQLException`, and we have higher-level exception handling that understands those. `RuntimeException` is a black box.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElementApi.java
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    Map<String, K8sPipelineElement> elements = new HashMap<>();
+    K8sPipelineElementStatusEstimator statusEstimator = new K8sPipelineElementStatusEstimator(context);
+    for (V1alpha1Pipeline pipeline : pipelines) {
+      List<K8sPipelineElementStatus> elementStatuses = statusEstimator.estimateStatuses(pipeline);
+      for (K8sPipelineElementStatus elementStatus : elementStatuses) {
+        String key = elementStatus.getName();
```
ryannedolan, line 37: The element's name is not necessarily a unique key. For example, you can have a `Pod` named `foo` and a `KafkaTopic` named `foo`, but those are separate things.

Since this is inside the `k8s` module, we can assume that these objects have all have a `kind`, at least. I think the key should at least be a tuple of name and kind.

Theoretically, you could also use the `uid` of each object, which is globally unique.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElementMapEntry.java
```diff
@@ -0,0 +1,20 @@
+package com.linkedin.hoptimator.k8s;
+
+/** Maps an element name to the name of one of its pipelines. */
+public class K8sPipelineElementMapEntry {
```
ryannedolan, line 4: Wondering if this needs to be public. Looks like it isn't used by any public methods.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElementMapTable.java
```diff
@@ -0,0 +1,41 @@
+package com.linkedin.hoptimator.k8s;
+
+import com.linkedin.hoptimator.util.RemoteTable;
+import org.apache.calcite.schema.Schema;
+
+
+/** Provides n:n mapping between pipeline elements and their pipelines. */
```
ryannedolan, line 7: nit, but I'd call this `n:m`

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElementMapTable.java
ryannedolan: This is super cool!

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineTable.java
```diff
@@ -11,16 +12,20 @@ public class K8sPipelineTable extends K8sTable<V1alpha1Pipeline, V1alpha1Pipelin
   // CHECKSTYLE:OFF
   public static class Row {
     public String NAME;
+    public boolean READY;
+    public boolean FAILED;
     public String STATUS;
```
ryannedolan, line 17: Wonder if we should rename `STATUS` to `MESSAGE`?

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElementApi.java
ryannedolan: Ah, I see from the integration tests that `getName()` is actually returning `Kind/name`. Works for me.

hoptimator-k8s/src/test/resources/k8s-metadata.id
```diff
@@ -0,0 +1,94 @@
...
+| ELEMENT_NAME                           | PIPELINE_NAME |
++----------------------------------------+---------------+
+| FlinkSessionJob/ads-database-audience2 | ads-audience2 |
+| FlinkSessionJob/ads-database-pages     | ads-pages     |
++----------------------------------------+---------------+
+(2 rows)
+
+!ok
+
+select pl.name as pipeline_name, pl.status as pipeline_status, pe.element_name, pe.failed as element_failed from "k8s".pipelines pl inner join (select t2.element_name, t1.failed, t2.pipeline_name from "k8s".pipeline_elements t1 inner join "k8s".pipeline_element_map t2 on t1.name = t2.element_name) pe on pl.name = pe.pipeline_name;
```
ryannedolan, line 85: This is amazing!

ryannedolan (approved):
🔥 🔥 🔥

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElement.java
nguyent8: Sure but just want to double check: I was trying to follow the model classes (e.g. V1alpha1Pipeline) where status() is for building, and there are get/setStatus(). wdyt?

ryannedolan: The data classes are auto-generated, so not really a conscience pattern. I don't have strong feelings either way. Both acceptable.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElementMapEntry.java
nguyent8: yeah converted this to default.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElementApi.java
nguyent8: Done

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElementMapTable.java
nguyent8: done

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineTable.java
nguyent8: sure. I also prefer message

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sPipelineElement.java
nguyent8: ah I see. I've updated this and others since they are not generated.

PR #137, opened 2025-05-12
Fix remaining spotbugs and checkstyle complaints and enable PR blocking

jogrogan:
- Fix remaining spotbugs and checkstyle complaints and enable PR blocking when either is broken.
- Enable Java 11

A lot of this was auto generated.
Verified existing tests work and cli functions normally

The standards and exclusions added are consistent with linkedin repositories.


ryannedolan (approved):
Nice! Funny to see the same mistakes made over and over.

PR #138, opened 2025-05-13
Add TableTriggers

ryannedolan:
## Summary

- Add `TableTrigger` CRD
- Add `TableTriggerReconciler`
- Add sample cron trigger

## Details

`TableTriggers` provide a mechanism to run arbitrary actions in response to table-level events, including inserts/updates/deletes. A trigger binds an action (encoded as a K8s Job template) to specific events for a specific table. When such events occur, the action is taken by deploying a Job.

`TableTriggers` can be fired by external systems. For demo purposes, a sample "cron trigger" is provided, which fires every minute.

## Testing Done

We can create a `TableTrigger`:

```
$ kubectl apply -f deploy/sample/tabletriggers.yaml
tabletrigger.hoptimator.linkedin.com/test-table-trigger created
```

The resulting trigger can be seen in the catalog:

```
0: Hoptimator> select * from "k8s".TABLE_TRIGGERS;
+--------------------+--------+------------------+-----------+-----------+
|        NAME        | SCHEMA |      TABLE       | TIMESTAMP | WATERMARK |
+--------------------+--------+------------------+-----------+-----------+
| test-table-trigger | KAFKA  | existing-topic-1 |           |           |
+--------------------+--------+------------------+-----------+-----------+
1 row selected (0.592 seconds)
```

To simulate a data change event, we can force the trigger to be "fired":

```
$ kubectl patch tabletrigger test-table-trigger -p "{\"status\":{\"timestamp\": \"`date +%FT%TZ`\"}}" --subresource=status --type=merge
```

We can observe a Job is deployed in response:

```
$ kubectl get jobs
NAME                     STATUS     COMPLETIONS   DURATION   AGE
test-table-trigger-job   Complete   1/1           46s        2m5s
```

After the job successfully runs, we see the trigger's watermark is updated and the job is deleted.

We can also create a "cron trigger" by driving our trigger from a cron job:

```
$ kubectl apply -f deploy/sample/crontrigger.yaml
```

We will then see that the trigger is fired every minute:

```
$ kubectl get tabletriggers
NAME                 SCHEMA   TABLE              TIMESTAMP                     WATERMARK
test-table-trigger   KAFKA    existing-topic-1   2025-05-13T18:48:00.000000Z   2025-05-13T18:47:00.000000Z
```


hoptimator-k8s/src/test/resources/k8s-metadata.id
```diff
@@ -49,18 +49,18 @@ select * from ads.pages;
...
 !ok
 
-select * from "k8s".pipelines;
-+---------------+-------+--------+---------+
-| NAME          | READY | FAILED | MESSAGE |
-+---------------+-------+--------+---------+
-| ads-audience2 | true  | false  | Ready.  |
-| ads-pages     | true  | false  | Ready.  |
-+---------------+-------+--------+---------+
+select name, failed from "k8s".pipelines order by name;
```
ryannedolan, line 52: @nguyent8 I relaxed these tests here, cuz they could get flakey. wdyt?

nguyent8: yes, this makes sense

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
```diff
@@ -0,0 +1,183 @@
...
+              .get(TRIGGER_TIMESTAMP_KEY) == null) {
+            log.error("Job {} has no timestamp annotation. Unable to progress the watermark.", name);
+          } else if (object.getStatus().getWatermark() == null) {
+            log.info("Trigger {} has no existing watermark.", name);
+            String watermark = job.getMetadata().getAnnotations().get(TRIGGER_TIMESTAMP_KEY);
+            status.setWatermark(OffsetDateTime.parse(watermark));
+            tableTriggerApi.updateStatus(object, status);
+            log.info("Trigger {} watermark set to {}.", name, watermark);
+          } else {
+            OffsetDateTime timestamp = OffsetDateTime.parse(job.getMetadata().getAnnotations().get(TRIGGER_TIMESTAMP_KEY));
```
srnand, line 123: this block's logic looks exactly similar to that of previous else if block's logic, do we need this ?

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
```diff
@@ -0,0 +1,183 @@
...
+            status.setWatermark(timestamp);
+            tableTriggerApi.updateStatus(object, status);
+            log.info("Trigger {} watermark advanced to {}.", name, timestamp);
+          }
+        }
+        if (failed || complete) {
```
srnand, line 129: could we add this logic in the previous if/else block itself ? so we dont need to have another conditional block ?
```
jobApi.delete(job);
return new Result(true); 
```

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
```diff
@@ -0,0 +1,183 @@
...
+      V1alpha1TableTriggerStatus status = object.getStatus();
+      if (status == null || status.getTimestamp() == null) {
+        log.info("Trigger has not been fired yet. Skipping.", name);
+        return new Result(false);
+      }
+
+      log.info("TableTrigger {} was last fired at {}.", name, status.getTimestamp());
+
+      if (object.getSpec().getYaml() == null) {
+        log.info("Trigger has no Job YAML. Will take no action.", name);
```
srnand, line 85: nit: missing string format specifiers ?

hoptimator-k8s/src/testFixtures/java/com/linkedin/hoptimator/k8s/FakeK8sYamlApi.java
```diff
@@ -0,0 +1,51 @@
...
+
+  @Override
+  public void createWithAnnotationsAndLabels(String yaml, Map<String, String> annotations,
+      Map<String, String> labels) throws SQLException {
+    DynamicKubernetesObject obj = Dynamics.newFromYaml(yaml);
+    obj.setMetadata(obj.getMetadata().annotations(annotations).labels(labels));
+    String dump = Yaml.dump(obj);
+    System.out.println("Created:");
+    System.out.println(dump);
+    System.out.println();
```
srnand, line 34: nit: should we remove these print statements?

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
```diff
@@ -0,0 +1,183 @@
...
+  } 
+
+  @Override
+  public Result reconcile(Request request) {
+    log.info("Reconciling request {}", request);
+    String name = request.getName();
+    String namespace = request.getNamespace();
+
+    try {
+      V1alpha1TableTrigger object = tableTriggerApi.get(namespace, name);
```
srnand, line 69: nit: maybe name this `tableTriggerObject` ?

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
```diff
@@ -0,0 +1,183 @@
...
+          } else {
+            OffsetDateTime timestamp = OffsetDateTime.parse(job.getMetadata().getAnnotations().get(TRIGGER_TIMESTAMP_KEY));
+            status.setWatermark(timestamp);
+            tableTriggerApi.updateStatus(object, status);
+            log.info("Trigger {} watermark advanced to {}.", name, timestamp);
+          }
+        }
+        if (failed || complete) {
+          jobApi.delete(job);
+          return new Result(true);  // start over
```
srnand, line 131: why do we return `Result(true)` here even when the job is finished ?

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
```diff
@@ -0,0 +1,183 @@
...
+import com.linkedin.hoptimator.k8s.K8sContext;
+import com.linkedin.hoptimator.k8s.K8sYamlApi;
+import com.linkedin.hoptimator.k8s.models.V1alpha1TableTrigger;
+import com.linkedin.hoptimator.k8s.models.V1alpha1TableTriggerList;
+import com.linkedin.hoptimator.k8s.models.V1alpha1TableTriggerStatus;
+import com.linkedin.hoptimator.util.Template;
+
+
+/**
+ * Launches Jobs when TableTriggers are fired.
```
nguyent8, line 35: could you expand on this so we can understand its behavior (esp. with watermarking) without reading the code?

hoptimator-k8s/src/testFixtures/java/com/linkedin/hoptimator/k8s/FakeK8sYamlApi.java
ryannedolan: This can only be used in tests. Let's keep, cuz they're useful to see in test logs.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
```diff
@@ -0,0 +1,183 @@
...
+      V1alpha1TableTrigger object = tableTriggerApi.get(namespace, name);
+
+      if (object == null) {
+        log.info("Object {} deleted. Skipping.", name);
+        return new Result(false);
+      }
+
+      V1alpha1TableTriggerStatus status = object.getStatus();
+      if (status == null || status.getTimestamp() == null) {
+        log.info("Trigger has not been fired yet. Skipping.", name);
```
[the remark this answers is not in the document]
ryannedolan: nice catch

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
ryannedolan: A reconciler only operates on one type of object, so this should be clear.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
ryannedolan: If we fire a trigger and there is already an existing failed/completed job, we must first delete the job before launching a new one. Rather than handle all these cases, we can consider the logic as having two main paths:

while we should fire:
1. if there is an existing job, wait for it to complete and then delete it
2. if there is no existing job, create a new one

The `Result(true)` implements the while-loop. I think the logic is simpler this way, but I could be wrong.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
ryannedolan: thanks, fixed

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
ryannedolan: done, thx

srnand (approved):
lgtm!

PR #142, opened 2025-05-22
Improve avro conversion and pass RelDataType through source/sinks

jogrogan:
It will be very common that deployers will actually need to know the schema of the data that they are working with.
The current K8sYamlDeployer doesn't, but other deployers will need to be involved in schema registration to the appropriate system.

This change adds the ability to load RelDataType of a source/sink dynamically via HoptimatorDriver. In practice this will cause the driver to load the same data multiple times, we likely want to improve this by caching drivers in some way such that repeated calls won't try to recreate the same connection.

This change also enhances the avro conversion capabilities from Calcite RelDataType and adds a helper method to divide fields into key and value schemas (some systems register both independently).

Also add a bunch of tests and fixed !specify and !pipeline commands which were broken with the introduction of the "pipeline" template option. Need to fill placeholders for this value or the rendering of the template will fail.

### DIFF
--- a/deploy/samples/etl-retl.yaml
+++ b/deploy/samples/etl-retl.yaml
@@ -1,0 +1,104 @@
+
+apiVersion: hoptimator.linkedin.com/v1alpha1
+kind: TableTemplate
+metadata:
+  name: kafka-source-trigger-template
+spec:
+  databases:
+  - kafka-database
+  methods:
+  - Scan
+  yaml: |
+    apiVersion: hoptimator.linkedin.com/v1alpha1
+    kind: TableTrigger
+    metadata:
+      name: {{name}}-trigger
+    spec:
+      schema: KAFKA
+      table: {{table}}
+      yaml: |
+        apiVersion: batch/v1
+        kind: Job
+        metadata:
+          name: {{name}}-job
+        spec:
+          template:
+            spec:
+              containers:
+              - name: hello
+                image: alpine/k8s:1.33.0
+                command: ["bash",  "-c", "echo {{name}}-trigger fired at `date`"]
+              restartPolicy: Never
+          backoffLimit: 4
+
+---
+
+## Simulate an upstream rETL job
+apiVersion: hoptimator.linkedin.com/v1alpha1
+kind: TableTemplate
+metadata:
+  name: kafka-source-cron-template
+spec:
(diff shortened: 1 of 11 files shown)


ryannedolan:
I get the motivation for this, but the fact that we sometimes know the schema and sometimes don't means this API is a bit janky.

There should be a way to get the RelDataType for any Source/Sink by resolving it from the catalog. Can deployers just do `HoptimatorDriver.resolve(source)` if they need to?

ryannedolan:
> > HoptimatorDriver
> 
> I believe in practice we should always know the schema.

That's good, but we're also leaking Calcite in the API now, which is a big dependency to ship around. I've been trying to limit Calcite to `hoptimator-utils` and `hoptimator-jdbc`. I'm thinking we should add `HoptimatorDriver.rowType(source)` (or something along those lines), like we do with `convert()`. Then plugins (including `Deployers`) can resolve the row type only if they need to.

Another approach would be to introduce our own `DataType` and pass that around in the API. Then `hoptimator-utils` could provide converters to/from `RelDataType`. That's probably more work than it's worth, but it would mean that our API doesn't depend on Calcite.

I've been bitten many times trying to ship Calcite around. Really would like to avoid that.

jogrogan:
> > > HoptimatorDriver
> > 
> > 
> > I believe in practice we should always know the schema.
> 
> That's good, but we're also leaking Calcite in the API now, which is a big dependency to ship around. I've been trying to limit Calcite to `hoptimator-utils` and `hoptimator-jdbc`. I'm thinking we should add `HoptimatorDriver.rowType(source)` (or something along those lines), like we do with `convert()`. Then plugins (including `Deployers`) can resolve the row type only if they need to.
> 
> Another approach would be to introduce our own `DataType` and pass that around in the API. Then `hoptimator-utils` could provide converters to/from `RelDataType`. That's probably more work than it's worth, but it would mean that our API doesn't depend on Calcite.
> 
> I've been bitten many times trying to ship Calcite around. Really would like to avoid that.

Agreed, will see if I can figure out something next week. After a quick look it doesn't seem like there is a nice way as schemas aren't stored in some singleton somewhere, they are a part of the connection/context objects that are passed around.

ryannedolan (approved):
Nice!

PR #144, opened 2025-05-29
Enable validating SqlCreateView and SqlCreateMaterializedView

nguyent8:
## Problem and Solution
Currently HoptimatorDdlExecutor only do validation in a later stage against the planned deployable objects created from validates SqlCreateView and SqlCreateMaterializedView.

The ValidationService doesn't validate create view commands directly.

This PR is to make ValidationService validate create view commands early. No validators are registered for those commands by default. They can be extended by users of hoptimator-jdbc library.

# Testing Done
Add new quidem test cases with a test validator to validating that create view commands are validated.

### DIFF
--- a/hoptimator-jdbc/src/test/java/com/linkedin/hoptimator/jdbc/TestSqlScripts.java
+++ b/hoptimator-jdbc/src/test/java/com/linkedin/hoptimator/jdbc/TestSqlScripts.java
@@ -1,6 +1,19 @@
 package com.linkedin.hoptimator.jdbc;
 
+import com.linkedin.hoptimator.Validator;
+import com.linkedin.hoptimator.ValidatorProvider;
+import java.io.IOException;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Collection;
+import java.util.List;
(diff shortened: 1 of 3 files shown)


hoptimator-jdbc/src/test/java/com/linkedin/hoptimator/jdbc/TestSqlScripts.java
```diff
@@ -9,4 +22,72 @@ public class TestSqlScripts extends QuidemTestBase {
...
+  }
+
+  private void useTestValidators() throws IOException {
+    Path tempDir = Files.createTempDirectory("spi-test");
+    Path servicesDir = tempDir.resolve("META-INF/services");
+    Files.createDirectories(servicesDir);
+    Files.writeString(servicesDir.resolve("com.linkedin.hoptimator.ValidatorProvider"),
+        "com.linkedin.hoptimator.jdbc.TestSqlScripts$CreateViewValidatorProvider\n");
+
+    URLClassLoader cl = new URLClassLoader(new URL[]{tempDir.toUri().toURL()}, getClass().getClassLoader());
```
ryannedolan, line 70: neat!

ryannedolan (approved):
Neat!

PR #149, opened 2025-07-21
Use TableTriggers As TableTemplates With Hint-able Expressions

srnand:
## Description
- Added logic to support table triggers as table templates within the source deployer. If the provided hints match the template variables, the corresponding templates are applied.
- For templates that cannot be resolved, exceptions are caught and a warning is logged without interrupting the process.
- To enable trigger creation during pipeline deployment, a custom hint with the key "offline.table.name" must be included in the query (via RPC).
- Since Calcite currently does not support inline SQL hints for DDL statements, this functionality is limited to queries executed through RPC.

## Testing
- Verified the end-to-end flow with targeted tests to ensure triggers are correctly deployed when the appropriate hint is provided.
- Confirmed that deleting the trigger resource disables it without impacting the pipeline's state.
- Validated that dropping a materialized view also removes any associated triggers, if present.

### *Request to create a new pipeline*
<p align="mid">
Request to create a new pipeline
</p>

### *Validating TableTrigger's deployment on that same source*

Validating TableTrigger's deployment on that same source


hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sTableTriggerDeployer.java
```diff
@@ -0,0 +1,90 @@
...
+import io.kubernetes.client.openapi.models.V1ObjectMeta;
+import io.kubernetes.client.openapi.models.V1PodSpec;
+import io.kubernetes.client.openapi.models.V1PodTemplateSpec;
+import io.kubernetes.client.util.Yaml;
+
+import java.sql.SQLException;
+import java.util.Collections;
+import java.util.Properties;
+
+class K8sTableTriggerDeployer extends K8sDeployer<V1alpha1TableTrigger, V1alpha1TableTriggerList> {
```
ryannedolan, line 18: It looks like this class doesn't do anything that a `TableTemplate` couldn't do? Can we just use a `TableTemplate`?

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sMaterializedViewDeployer.java
```diff
@@ -61,6 +74,15 @@ public void update() throws SQLException {
       deployers.add(pipelineDeployer);
       V1OwnerReference pipelineRef = pipelineDeployer.updateAndReference();
       K8sContext pipelineContext = viewContext.withLabel("pipeline", name).withOwner(pipelineRef);
+      // Only update a TableTrigger if there is exactly one source table and if the OFFLINE_TABLE_NAME hint is set.
+      String offlineTableName = hints.get(OFFLINE_TABLE_NAME);
+      if (view.pipeline().sources().size() == 1 && !Strings.isNullOrEmpty(offlineTableName)) {
```
ryannedolan, line 79: I guess eventually hints will work for multiple sources; e.g.:

```
create materialized view foo as
   select * from bar /*+ hint('offline.table.name'='BAR') */,
   qux /*+ hint('offline.table.name'='QUX') */
```

ryannedolan (commented):
hmm, I think maybe this logic belongs in `K8sSourceDeployer` somehow. Ideally we'd have a way to enable/disable any `TableTemplate`, rather than have logic specifically to enable/disable `TableTriggers`.

My thinking is that we need a way for `TableTemplates` to define their own required hints. Then we'd just need a `TableTemplate` for `TableTriggers`, which only applies when the `offline.table.name` hint is present.

Two ideas:

1. We add a `hints` section to `TableTemplate` CRD. In order for the template to apply, the current hints must match the values specified in `.spec.hints`.
2. We just ignore templates that have missing variables, e.g. `{{offline.table.name}}`. Rather than throw an exception ("missing variable" or whatever), we just ignore the template.

Slight preference for (2) just cuz we wouldn't need to change the CRD.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sMaterializedViewDeployer.java
ryannedolan: Yeah, it will take a lot of work to get these hints working within our DDL, but it is possible. We've looked into it multiple times, but haven't prioritized it yet. In the meantime, we can enable hints on the entire connection.

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sSourceDeployer.java
```diff
@@ -6,34 +6,39 @@
...
+    HoptimatorConnection connection = context.connection();
     Template.Environment env =
         new Template.SimpleEnvironment()
             .with("name", name)
             .with("database", source.database())
             .with("schema", source.schema())
             .with("table", source.table())
-            .with(source.options());
+            .with(source.options())
+            .with(DeploymentService.parseHints(connection.connectionProperties()));
```
ryannedolan, line 40: neat

hoptimator
```diff
@@ -8,5 +8,5 @@ $BASEDIR/hoptimator-cli/build/install/hoptimator-cli/bin/hoptimator-cli \
   sqlline.SqlLine \
   -ac sqlline.HoptimatorAppConfig \
   --verbose \
-  -u "jdbc:hoptimator://fun=mysql" -n "" -p "" -nn "Hoptimator" $@
+  -u "jdbc:hoptimator://fun=mysql;hints=offline.table.name=ads_offline" -n "" -p "" -nn "Hoptimator" $@
```
ryannedolan, line 11: I guess we need to get hints working in the CLI somehow.

Maybe we should add a `!hint foo bar` command to the CLI that invokes `HoptimatorConnection.setHint(foo, bar)`. I think we may need to refactor how hints are passed around tho.

ryannedolan: (This is fine for now.)

ryannedolan (approved):
lgtm!

PR #150, opened 2025-07-22
Clean up several subscription artifacts

jogrogan:
1. Fix issue in AvroConverter caused by change in #148 
2. Remove old Subscription model files & `HoptimatorOperatorApp` which is no longer in use.
3. Fix gaps in `PipelineOperatorApp` - namely around `ControllerProvider` which was replaced with new one based on `K8sContext` instead of `Operator`. This was preventing extensions to add new controllers.


hoptimator-kafka-controller/src/main/java/com/linkedin/hoptimator/operator/kafka/KafkaTopicReconciler.java
```diff
@@ -19,22 +24,25 @@
...
   private static final Logger log = LoggerFactory.getLogger(KafkaTopicReconciler.class);
-  private static final String KAFKATOPIC = "hoptimator.linkedin.com/v1alpha1/KafkaTopic";
 
-  private final Operator operator;
+  private final K8sContext context;
+  private final K8sApi<V1alpha1KafkaTopic, V1alpha1KafkaTopicList> kafkaTopicApi;
 
-  public KafkaTopicReconciler(Operator operator) {
-    this.operator = operator;
+  public KafkaTopicReconciler(K8sContext context) {
```
ryannedolan, line 39: nice!

ryannedolan (approved):

PR #151, opened 2025-07-29
Add ResolvedTable

ryannedolan:
## Summary

- Added `ResolvedTable` class.
- Added `HoptimatorConnection.resolve()` method.
- Added `!resolve foo.bar` CLI command.

## Details

In order to use Hoptimator's table resolution machinery in external catalog implementations (e.g. a Flink catalog), we need the ability to resolve table schemas / row types and connector configurations over a JDBC connection. For convenience purposes, this PR pulls together such metadata into a new `ResolvedTable` class, including an Avro-encoded schema.

Connector configurations are left as a TODO, since this machinery is still a work in progress elsewhere.

A new `!resolve` CLI command dumps this new metadata.

## Testing done

```
: Hoptimator> !resolve ADS.AD_CLICKS
Avro schema:

{
  "type" : "record",
  "name" : "ad_clicks",
  "namespace" : "com.linkedin.ads",
  "doc" : "RecordType(VARCHAR CAMPAIGN_URN, VARCHAR MEMBER_URN)",
  "fields" : [ {
    "name" : "CAMPAIGN_URN",
    "type" : [ "null", "string" ],
    "doc" : "CAMPAIGN_URN VARCHAR"
  }, {
    "name" : "MEMBER_URN",
    "type" : [ "null", "string" ],
    "doc" : "MEMBER_URN VARCHAR"
  } ]
}
```

### DIFF
--- a/hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
+++ b/hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
@@ -4,7 +4,10 @@ import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Locale;
+import java.util.Map;
 import java.util.Properties;
 import java.util.stream.Collectors;
 
@@ -13,6 +16,7 @@ import org.apache.calcite.jdbc.CalcitePrepare;
 import org.apache.calcite.plan.RelOptMaterialization;
 import org.apache.calcite.rel.RelNode;
 
+import com.linkedin.hoptimator.avro.AvroConverter;
 import com.linkedin.hoptimator.util.DelegatingConnection;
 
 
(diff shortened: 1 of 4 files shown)


hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
```diff
@@ -28,6 +32,26 @@ public HoptimatorConnection(CalciteConnection connection, Properties connectionP
...
+      RelNode tableRel = HoptimatorDriver.convert(this, tableSql).root.rel;
+      String namespace = "com.linkedin."
+            + tablePath.stream()
+            .map(x -> x.toLowerCase(Locale.ROOT))
+            .limit(tablePath.size() - 1)
+            .collect(Collectors.joining("."));
+      String schemaName = tablePath.get(tablePath.size() - 1).toLowerCase(Locale.ROOT);
+      org.apache.avro.Schema avroSchema = AvroConverter.avro(namespace, schemaName, tableRel.getRowType());
+      // TODO: generate source and sink configs via ConnectionService
+      return new ResolvedTable(tablePath, avroSchema, Collections.emptyMap(), Collections.emptyMap());
```
jogrogan, line 49: Is the intention to provide the configs as if the tablePath would be used as a source or sink? And then whoever is using the table resolver will select the one they need?

ryannedolan: Yes. From the perspective of the data plane (e.g. a Flink catalog) we don't know how a table will be used until it is used. So we resolve for both situations, and the data plane can choose one or both as needed.

The connector configs may use the same or different physical connectors. For example, table templates could construct a table that reads with Iceberg and writes with Kafka. Theoretically, the source connector may itself be "hybrid", e.g. it could start reading from Iceberg and cutover to Kafka.

jogrogan: Sounds reasonable to me. Should be pretty easy to add the ConnectionService call too. Might get a little awkward because some schemas can be used as a source and not a sink and maybe vice versa but as long as it's obvious to the user (empty map or null?) it should be fine.

PR #154, opened 2025-07-29
Add logging function to HoptimatorConnection so its clients can retrieve logs by registering hooks

nguyent8:
**Problem:**
Currently Hoptimator jdbc doesn't provide execution details such as validations or deployment steps other than the number of modified rows. Some clients are interested in those details to help users understand what happen behind the scene.

**Solution:**
This PR add logging function to HoptimatorConnection for create (materialized) view statements so its clients can retrieve logs at HoptimatorDdlExecutor level by registering hooks. It has a pending TODO to add statement execution log isolation in the future.

Logging at lower level than the HoptimatorDdlExecutor could be added in another PR if desired.

With this PR, when executing "create materialized view \"QUEUING\".test_view_00 as select * from \"QUEUING\".\"test-source\"" statement, hoptimator client would get the following log lines:
```
"[HoptimatorDdlExecutor] Validating statement: CREATE MATERIALIZED VIEW `QUEUING`.`TEST_VIEW_00` AS\nSELECT *\nFROM `QUEUING`.`test-source`",
"[HoptimatorDdlExecutor] Validated sql statement. The view is named TEST_VIEW_00 and has path [QUEUING, TEST_VIEW_00]",
"[HoptimatorDdlExecutor] Pipeline name for view TEST_VIEW_00 is kafka-queuing-TEST_VIEW_00",
"[HoptimatorDdlExecutor] Added view TEST_VIEW_00 to schema QUEUING",
"[HoptimatorDdlExecutor] Validating view TEST_VIEW_00 with deployers",
"[HoptimatorDdlExecutor] Validated view TEST_VIEW_00",
"[HoptimatorDdlExecutor] Deploying view TEST_VIEW_00",
"[HoptimatorDdlExecutor] Deployed view TEST_VIEW_00",
"[HoptimatorDdlExecutor] CREATE MATERIALIZED VIEW TEST_VIEW_00 completed"
```

**Testing Done:**

- Add new unit test: com.linkedin.hoptimator.jdbc.TestBasicSql.createView
- Manual run a hoptimator jdbc client and verify logs

### DIFF
--- a/hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
+++ b/hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
@@ -32,7 +32,7 @@ public class HoptimatorConnection extends DelegatingConnection {
   private final Properties connectionProperties;
   private final List<RelOptMaterialization> materializations = new ArrayList<>();
 
-  private final List<Consumer<String>> hooks = new ArrayList<>();
+  private final List<Consumer<String>> logHooks = new ArrayList<>();
 
   public HoptimatorConnection(CalciteConnection connection, Properties connectionProperties) {
     super(connection);
@@ -100,7 +100,7 @@ public class HoptimatorConnection extends DelegatingConnection {
    * Returns a logger for a client of this connection. The logger logs to both SLF4J and hooks.
    */
   HoptimatorConnectionDualLogger getLogger(Class<?> clazz) {
-    return new HoptimatorConnectionDualLogger(clazz, hooks);
+    return new HoptimatorConnectionDualLogger(clazz, logHooks);
   }
 
   /**
@@ -108,7 +108,7 @@ public class HoptimatorConnection extends DelegatingConnection {
    * TODO: Revise to allow hooks to be added per statement.
    */
   public void addLogHook(Consumer<String> hook) {
-    hooks.add(hook);
+    logHooks.add(hook);
   }
 
   private void registerMaterialization(List<String> viewPath, RelNode tableRel, RelNode queryRel) {
--- a/hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
+++ b/hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
@@ -3,31 +3,27 @@ package com.linkedin.hoptimator.jdbc;
 import com.linkedin.hoptimator.Database;
 import com.linkedin.hoptimator.Sink;
 import com.linkedin.hoptimator.Source;
+import com.linkedin.hoptimator.avro.AvroConverter;
 import com.linkedin.hoptimator.util.ConnectionService;
+import com.linkedin.hoptimator.util.DelegatingConnection;
 import java.sql.PreparedStatement;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Properties;
+import java.util.function.Consumer;
 import java.util.stream.Collectors;
-
 import org.apache.avro.Schema;
 import org.apache.calcite.jdbc.CalciteConnection;
 import org.apache.calcite.jdbc.CalcitePrepare;
 import org.apache.calcite.jdbc.CalciteSchema;
 import org.apache.calcite.plan.RelOptMaterialization;
 import org.apache.calcite.rel.RelNode;
-
-import com.linkedin.hoptimator.avro.AvroConverter;
-import com.linkedin.hoptimator.util.DelegatingConnection;
 import org.apache.calcite.util.Util;
-import org.slf4j.Logger;
 
 
 public class HoptimatorConnection extends DelegatingConnection {
@@ -36,27 +32,7 @@ public class HoptimatorConnection extends DelegatingConnection {
   private final Properties connectionProperties;
   private final List<RelOptMaterialization> materializations = new ArrayList<>();
 
-  private final Map<Class<?>, HoptimatorConnectionDualLogger> loggers = new HashMap<>();
-
-  public HoptimatorConnectionDualLogger getLogger(Class<?> clazz) {
-    return loggers.computeIfAbsent(clazz, k -> new HoptimatorConnectionDualLogger(clazz));
-  }
-
-  public List<String> getLogs() {
-    return getLogsByClass().entrySet()
-        .stream()
-        .map(e -> e.getKey().getSimpleName() + ": " + String.join("\n", e.getValue()))
-        .collect(Collectors.toList());
-  }
-
-  public Map<Class<?>, List<String>> getLogsByClass() {
-    return loggers.entrySet().stream().collect(Collectors.toMap(Map.Entry::getKey, e -> e.getValue().getLogs()));
-  }
-
-  public void clearLogs() {
-    loggers.clear();
-  }
-
+  private final List<Consumer<String>> hooks = new ArrayList<>();
 
   public HoptimatorConnection(CalciteConnection connection, Properties connectionProperties) {
     super(connection);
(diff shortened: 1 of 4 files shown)


hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
```diff
@@ -22,6 +25,28 @@ public class HoptimatorConnection extends DelegatingConnection {
...
+  private final Map<Class<?>, HoptimatorConnectionDualLogger> loggers = new HashMap<>();
+
+  public HoptimatorConnectionDualLogger getLogger(Class<?> clazz) {
+    return loggers.computeIfAbsent(clazz, k -> new HoptimatorConnectionDualLogger(clazz));
+  }
+
+  public List<String> getLogs() {
+    return getLogsByClass().entrySet()
+        .stream()
+        .map(e -> e.getKey().getSimpleName() + ": " + String.join("\n", e.getValue()))
```
nguyent8, line 37: flatten?
classname: logline1
classname: logline2

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
```diff
@@ -22,6 +25,28 @@ public class HoptimatorConnection extends DelegatingConnection {
   private final Properties connectionProperties;
   private final List<RelOptMaterialization> materializations = new ArrayList<>();
 
+  private final Map<Class<?>, HoptimatorConnectionDualLogger> loggers = new HashMap<>();
```
nguyent8, line 28: We can find a way to associate an uid w/ a query to have better log isolation.
However we will need to explicitly state that HoptimatorConnection is not thread-safe. HoptimatorConnection actually doesn't look thread-safe before this PP either.

nguyent8: It seems: to associate uid provided by clients with a sql statement, we will need to extend SQL grammar and implement SqlParser which parse new statement which could be a modification of createViewStatement with new hint info similar to the hintComment.

ryannedolan: It could be done with a metadata table I guess; e.g. `select msg from metadata.logs where statementId = 123 and logLevel = 'WARN'`

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
```diff
@@ -22,6 +25,28 @@ public class HoptimatorConnection extends DelegatingConnection {
...
+
+  public HoptimatorConnectionDualLogger getLogger(Class<?> clazz) {
```
ryannedolan, line 30: Probably want the return type to be just `java.util.logging.Logger` to avoid leaking this internal class.

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
```diff
@@ -68,4 +93,24 @@ private void registerMaterialization(List<String> viewPath, RelNode tableRel, Re
   public List<RelOptMaterialization> materializations() {
     return materializations;
   }
+
+  /** A logger that logs to both SLF4J and an internal list. */
+  public static class HoptimatorConnectionDualLogger {
```
ryannedolan, line 98: Can probably make this private?

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
```diff
@@ -22,6 +25,28 @@ public class HoptimatorConnection extends DelegatingConnection {
   private final Properties connectionProperties;
   private final List<RelOptMaterialization> materializations = new ArrayList<>();
 
+  private final Map<Class<?>, HoptimatorConnectionDualLogger> loggers = new HashMap<>();
+
+  public HoptimatorConnectionDualLogger getLogger(Class<?> clazz) {
+    return loggers.computeIfAbsent(clazz, k -> new HoptimatorConnectionDualLogger(clazz));
+  }
+
+  public List<String> getLogs() {
```
ryannedolan, line 34: Instead of accumulating logs in a map like this, can the caller just register a hook? I was thinking along these lines:

```
Connection conn = DriverManager.connect("jdbc:hoptimator://", props);
conn.setInfoLogHook(x -> response.onNext(Response.newBuilder().addInfoLogLine(x));
conn.setErrorLogHook(x -> response.onNext(Response.newBuilder().addErrorLogLine(x));
...
```

or smth like that, which would let us stream loglines to the client in a gRPC server.

In our non-streaming endpoints (`Query`, `Execute`) we could just add the log messages directly to the response builder:

```
conn.setInfoLogHook(x -> responseBuilder.addInfoLogLine(x));
...
onNext(responseBuilder.build());
```

In either case, we don't really need `HoptimatorConnection` to store the logs. We just need it to keep track of some optional callbacks/hooks.

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
```diff
@@ -68,4 +93,24 @@ private void registerMaterialization(List<String> viewPath, RelNode tableRel, Re
...
+    private final Logger slf4jLogger;
+
+    public HoptimatorConnectionDualLogger(Class<?> clazz) {
+      this.slf4jLogger = org.slf4j.LoggerFactory.getLogger(clazz);
+    }
+
+    private final List<String> logs = new ArrayList<>();
+
+    public void info(String format, Object... arguments) {
+      logs.add(String.format(format, arguments));
```
ryannedolan, line 108: See above wrt hooks. Instead of `logs.add()` here, you could just invoke the hook: `infoLogHook.accept(msg)`.

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
nguyent8: per our discussion, we will move forward without log isolation per statement. Added a todo to revisit this later.

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
nguyent8: I considered that but chose this custom class which has package scope, mainly to:
- avoid having to override a very long list of methods on java.util.logging.Logger
- avoid the incompatibility in string formatting between the standard and slf4j styles

Generally I think this short class is more comprehensible.
But do lmk if you feel strong about exposing a java.util.logging.Logger instead.

jogrogan (approved):
Approving, mostly nits. Let me know what you think

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorConnection.java
```diff
@@ -33,6 +32,8 @@ public class HoptimatorConnection extends DelegatingConnection {
   private final Properties connectionProperties;
   private final List<RelOptMaterialization> materializations = new ArrayList<>();
 
+  private final List<Consumer<String>> hooks = new ArrayList<>();
```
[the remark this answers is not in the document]
nguyent8: definnitely a better name

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorDdlExecutor.java
```diff
@@ -133,18 +136,26 @@ public void execute(SqlCreateView create, CalcitePrepare.Context context) {
     RelProtoDataType protoType = RelDataTypeImpl.proto(analyzed.rowType);
     ViewTable viewTable = new ViewTable(Object.class, protoType, sql, schemaPath, viewPath);
     View view = new View(viewPath, sql);
+    logger.info("Validated sql statement. The view is named {} and has path {}",
+        viewName, viewPath);
 
     Collection<Deployer> deployers = null;
     try {
+      logger.info("Validating view {} with deployers", viewName);
```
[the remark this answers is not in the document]
nguyent8: yeah agree. I thought the same but wanted to keep this PR at this level for now. We definitely need to go to lower levels (mentioned in the description)

PR #156, opened 2025-08-01
Fix nested array conversion

skolasan:
RelDataType to avro conversion fails for nested Array of Structs due to null being passed to sanitize function. This change fixes the issue by adding a default recordName for array and map types


ryannedolan:
Thanks!

PR #181, opened 2025-12-10
Fix loading namespace from kubectl

ryannedolan:
## Summary

- Fixed incorrect loading of namespace from kubeconfig.

## Details

Unless explicitly configured to use a specific namespace, `K8sContext` (and thus all of the driver) was defaulting to `default` namespace, rather than using the current namespace in `/.kube/config`.

The behavior of the driver when run within a K8s pod isn't changed.

## Testing

Spot checked with `hoptimator-cli`.


jogrogan:
Closing some old PRs - don't think this is an issue anymore

PR #206, opened 2026-04-13
Rewrite SubscriptionReconciler to use new planner and Pipeline CRs

ryannedolan:
## Summary
- Rewrote `SubscriptionReconciler` to use `DeploymentService.plan()` / `PipelineRel` instead of the legacy `HoptimatorPlanner`, `Operator`, and `Resource` template system
- Reconciler now creates `Pipeline` K8s CRs owned by Subscriptions, delegating readiness monitoring to the existing `PipelineReconciler`
- Registered `SUBSCRIPTIONS` endpoint in `K8sApiEndpoints` and wired the controller into `PipelineOperatorApp`
- Removed `SubscriptionEnvironment` (deprecated) and dropped `hoptimator-planner`/`hoptimator-catalog`/`hoptimator-avro` dependencies from the operator module
- Planning is abstracted via a `Planner` interface for testability, with `jdbcPlanner()` factory for production use
- Fixed a hanging `K8sJobDeployerTest` caused by stale mock stubs

**Migration note:** The `hoptimator-planner` module is intentionally untouched — the internal `SubscriptionOperator` fork still depends on the published jar. This PR provides a clear upgrade path: deploy the new operator and existing `Subscription` CRs will be handled by the new planner.

## Testing Done
- [x] Unit tests rewritten for `SubscriptionReconciler` using `FakeK8sApi` pattern
- [x] All existing operator, k8s, jdbc, util, and api tests pass
- [x] Fixed previously-hanging `K8sJobDeployerTest`
- [x] Full `./gradlew test` passes with 0 failures


hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/PipelineOperatorApp.java
```diff
@@ -59,25 +62,42 @@ public static void main(String[] args) throws Exception {
...
     connectionProperties.put("k8s.watch.namespace", watchNamespaceInput);
-    K8sContext context = K8sContext.create(new HoptimatorConnection(null, connectionProperties));
-    new PipelineOperatorApp(context).run();
+
+    // Create a JDBC connection for SQL planning (used by SubscriptionReconciler)
+    Connection jdbcConnection = DriverManager.getConnection("jdbc:hoptimator://", connectionProperties);
+    K8sContext context = K8sContext.create(jdbcConnection);
+
+    SubscriptionReconciler.Planner planner = SubscriptionReconciler.jdbcPlanner(jdbcConnection);
+    new PipelineOperatorApp(context).run(planner);
```
jogrogan, line 71: It seems like there is a lot of logic just to allow SubscriptionReconciler to support drop in planners, though in practice it'll always use the jdbcPlanner. Do you think we need this?

hoptimator-operator/build.gradle
```diff
@@ -6,9 +6,6 @@ plugins {
```
jogrogan, line 6: I suspect there are more files that are now unused in the hoptimator-operator package, namely "Operator.java" but maybe more that we can remove.

jogrogan (commented):
Change overall makes sense. Discussed already, but think we need some additional testing with our internal extensions before we can merge

PR #209, opened 2026-04-20
Added Date and Timestamp Support in Avro's To/From Converter

srnand:
## Summary
  - Add Calcite `DATE` and `TIMESTAMP` type support to `AvroConverter` using Avro logical types.
  - Added support for both directions, `avro()` (Calcite → Avro) and `rel()` (Avro → Calcite).

  ## Testing
  - Added tests for both DATE and Timestamp to validate the conversion of Calcite type → Avro type → Calcite type.


ryannedolan (approved):
nice

PR #210, opened 2026-04-21
Lazy schema loading: defer implicit sub-schema discovery and stabilize for connection lifetime

jogrogan:
### New: HoptimatorCalciteSchema
A CalciteSchema subclass (required because SimpleCalciteSchema is package-private) that overrides createSubSchema() to pin each discovered implicit sub-schema into subSchemaMap on first access. Without this, Calcite creates a fresh wrapper on every subSchemas().get(name) call — meaning views registered on one wrapper are invisible to the next call. Pinning makes implicit sub-schemas (e.g. MySQL databases under a catalog) behave identically to explicit ones for the full connection lifetime. Previously we worked around this by force registering all schemas within K8sDatabaseTable

### New: MySqlRootSchema
A backing schema for MySQL that exposes its databases as lazy implicit sub-schemas. Previously, connecting to MySQL via the Calcite JDBC driver triggered getSchemas() on every metadata call, enumerating all databases. MySqlRootSchema defers this until a database is actually accessed.

### New: LazyLookup<T> (replaces LazyTableLookup)
Renamed and generalized from table-specific to generic — same lazy load-on-demand semantics (load(name) for targeted lookups, loadAll() for full enumeration) but applicable to schemas, tables, or any entity. All callers updated (DatabaseSchema, K8sMetadata, ClusterSchema, etc.) with normalized method names.

### Modified: CalciteDriver
Added three overridable hooks: createRootSchema(properties) to supply a custom root schema, onConnectionInit() to skip Calcite's eager lattice scan, and includeMetadataSchema() to opt into metadata.COLUMNS / metadata.TABLES.

## Important Note
While this change **does** defer loading of all schemas at connection start up time, Calcite still loads all schemas on any use of this driver. https://github.com/apache/calcite/pull/4891 intends to fix this problem, will need to eventually update Calcite versions once released.

### Test plan
- Added a lot of unit tests
- Verified lazy loading of MySQL schemas on connection startup


ryannedolan (approved):

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorCalciteSchema.java
```diff
@@ -0,0 +1,274 @@
...
+   * re-invoking this method.
+   *
+   * <p>If a schema with this name already exists in {@code subSchemaMap} (due to a prior
+   * {@code add()} call or a previous invocation of this method), the existing entry is
+   * returned so that any tables already registered to it remain visible.
+   */
+  @Override
+  protected CalciteSchema createSubSchema(Schema schema, String name) {
+    NavigableMap<String, CalciteSchema> existing = subSchemaMap.range(name, false);
+    if (!existing.isEmpty()) {
```
srnand: not sure but does this needs to be thread safe ? "checking and putting in a map" ?

srnand (approved):

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorCalciteSchema.java
jogrogan: Hmm good question, I did some digging and everything I'm seeing is that these structures in Calcite are not thread safe _however_ Calcite's design implicitly assumes subSchemaMap is mutated only at schema-setup time (explicit add() calls before any queries run), never during lookups so Calcite won't run into this problem. I think it makes sense to add a lock here, good call

PR #215, opened 2026-04-28
Add CREATE JOB DDL

ryannedolan:
## Summary

- Added `create job` ddl.

## Details

Adds `create [or replace] [flink] [streaming | batch] job as 'insert into ... ' [with (... options ...)]` statement. This enables automation to bypass Hoptimator's pipeline planner and create Flink jobs directly. This is particularly useful if the target Flink environment includes a pre-configured catalog, since in that case jobs can be simple one-liners.

## Testing Done

```
> create flink streaming job "my-etl-job" as 'INSERT INTO sink SELECT * FROM source' WITH (checkpointing.interval '10000');
```
produces:
```
kind: SqlJob
metadata:
  name: my-etl-job
spec:
  configs:
    checkpointing.interval: "10000"
  dialect: Flink
  executionMode: Streaming
  sql:
  - INSERT INTO sink SELECT * FROM source
  ```

### CODE REVIEW

**Comment by ryannedolan** [comment]
@copilot please resolve merge conflicts

**Comment by ryannedolan** [comment]
@copilot please add unit tests similar to #214

File: hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorDdlUtils.java [hunk 1]
```diff
@@ -672,6 +674,69 @@ static SpecifyResult processCreateDatabase(HoptimatorConnection conn,
...
+
+    if (create.name.names.size() > 1) {
+      throw new SQLException("Job names cannot be compound identifiers.");
+    }
+    String name = create.name.names.get(0);
+    String sqlBody = ((SqlLiteral) create.sqlBody).getValueAs(String.class);
+
+    // Split SQL body on semicolons into individual statements
+    List<String> sqlStatements = new ArrayList<>();
+    for (String stmt : sqlBody.split(";")) {
```
srnand, line 702: guess this split(";") may create edge-case conditions for any ';' character inside a Flink connector option value. noticed this since we saw something similar here, https://github.com/linkedin/Hoptimator/pull/199/changes

hoptimator-api/src/main/java/com/linkedin/hoptimator/SqlJobDeployable.java
```diff
@@ -0,0 +1,51 @@
+package com.linkedin.hoptimator;
+
+import java.util.List;
+import java.util.Map;
+
+
+/** Represents a CREATE JOB request for deploying a SqlJob to Kubernetes. */
+public class SqlJobDeployable implements Deployable {
```
jogrogan, line 8: More so naming wise but does feel awkward that we now have `SqlJobDeployable` and `Job` (which deploys SqlJobs via templates) implementing the same interface but do semi-different things - not a deal breaker

jogrogan (commented):
Fundamentally I am okay with this change, however it does break a lot of the "confidence" checks we get using our existing machinery. This includes all of our custom validators but also everything we get for free via the Calcite planner like validating syntax and fields. There's no way to guarantee that the specified SQL will actually work in Flink (or whatever engine).

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorDdlUtils.java
ryannedolan: good call will fix

PR #218, opened 2026-04-30
Expose non-blocking start()/stop() on PipelineOperatorApp + fail-fast failure reporting

jogrogan:
## Summary
                                                                                                                                                                                               
Restructures `PipelineOperatorApp` to support hosting under an external lifecycle (Spring,
Guava, etc.) without the caller having to spin up its own thread
to escape `ControllerManager.run()`'s indefinite `CountDownLatch.await()`. Adds explicit
detection for unexpected controller termination so callers can wire it to a liveness probe
and let the orchestrator handle restart/                                                                                                                                                     
                                                                                                                                                                                               
## Motivation                                                                                                                                                                                  
                                                                
The previous `run()` / `run(List<Controller>)` API blocks until shutdown because                                                                                                               
`ControllerManager.run()` parks on a `CountDownLatch`. That works fine for `main()` but
breaks any caller that owns the calling thread for lifecycle purposes. Concretely, hosting                                                                                                     
this in a Spring Boot `CommandLineRunner` either wedges the lifecycle thread or requires
the caller to manually offload the call.                                                                                                
                                                                                                                                                                                               
A second, related gap: when a controller's `run()` returns unexpectedly ("all" or "some"
controllers exiting silently), the manager catches the exit, logs, and                                                                                                      
keeps blocking. Nothing surfaces the failure to the caller. Frameworks like fabric8's                                                                                                          
java-operator-sdk and controller-runtime expose this so the orchestrator can restart the                                                                                                       
process; this PR adds the same hook here.
(Though maybe we should just migrate to fabric8 at some point?)
                                                                                                                                                                                               
## Backward compatibility                                                                                                                                                                         
 
This is a breaking change — run() and run(List<Controller>) are removed. The                                                                                                                   
in-tree main() was the only caller within this repo and is updated.                                                                                                                                
                                                                                                                                                                                               
## Testing
- Unit tests pass
- Stood up hoptimator-operator pod via `make deploy` and tested early termination cleanly shuts down the pod.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -12,9 +12,6 @@ ext.jacocoAggregateExcludes = [
   // ── hoptimator-models / hoptimator-k8s ──────────────────────────────────────
   '**/models/**',           // auto-generated K8s CRD model classes
 
-  // ── hoptimator-operator ──────────────────────────────────────────────────────
-  '**/PipelineOperatorApp.class',                  // operator main entry point
-
   // ── hoptimator-flink-runner ──────────────────────────────────────────────────
   '**/FlinkRunner.class',                          // CLI entry point; requires Flink runtime
 
--- a/deploy/dev/kafka.yaml
+++ b/deploy/dev/kafka.yaml
@@ -15,7 +15,7 @@
 # Based on examples at:
 # https://github.com/strimzi/strimzi-kafka-operator/blob/main/examples/kafka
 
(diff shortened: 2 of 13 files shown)


ryannedolan (approved):
https://github.com/linkedin/Hoptimator/pull/218/changes

PR #221, opened 2026-05-12
Extend pre-delete dep-guard to triggers

jogrogan:
## Summary

Extends the pre-delete dependency guard (which previously only saw Pipeline CRDs) to also block deletes when a TableTrigger references the resource being dropped. Along the way it generalizes the labels/checker machinery so it's not
Pipeline-specific in name or shape.

Stacked on jogrogan/dependency-guard-validator-v2.

## What's in it

- Split the dep annotation into directional depends-on-sources and depends-on-sinks. The old single depends-on annotation lost source/sink direction, which the visualizer (next PR up the stack) needs and the checker was already implicitly reading as two-sided.
- Dep-guard scans TableTrigger CRDs. DependencyChecker.findBlockers is generic over any KubernetesObject; we just call it twice (Pipelines and TableTriggers) and merge the blocker lists. Self-owned children are excluded from both sides so LogicalTableDeployer.delete() can still cascade.
- Renames — PipelineDependency{Labels,Checker} → Dependency{Labels,Checker}, K8sPipelineDependencyValidator → K8sDependencyValidator. Nothing about these is Pipeline-specific anymore.
- Collapsed labels API — single DependencyLabels.stamp(meta, Collection<Source>, Collection<Sink>) that writes both the labels and the directional annotations. Replaces three separate labelsFor/sourcesAnnotation/sinkAnnotation helpers and the
inline duplication that had crept into K8sTriggerDeployer.
- Wire format is symmetric — both depends-on-sources and depends-on-sinks are comma-separated identifier lists. Today every caller passes a singleton sink, but the format doesn't bake in the assumption — a future multi-sink resource needs no migration.
- Adds optional Sink to trigger - currently only used by LogicalTables for a known rETL trigger, this should be generalized (TODO left in code).

## Test plan
- K8sTriggerDeployerTest — adds coverage for the dep-label stamping on both update (pause/resume partial path) and toK8sObject (full path), for source-only and source+sink trigger shapes.
- k8s-trigger-validation.id Quidem integration test: DROP TABLE on a source with an active trigger is blocked with the expected error message.


ryannedolan:
@copilot resolve the merge conflicts in this pull request

ryannedolan (approved):
neat!

PR #224, opened 2026-05-14
Infer path variable from ON-clause path

srnand:
## Summary
- Replaced `{{offline.table.name}}` in JobTemplates with `{{path}}` which now defaults to `source.pathString()` (the fully-qualified `CATALOG.SCHEMA.TABLE` / `SCHEMA.TABLE` from the trigger's ON clause), so users no longer need to repeat the path in `WITH ('offline.table.name' '...')`.

### DIFF
--- a/hoptimator
+++ b/hoptimator
@@ -8,5 +8,5 @@ $BASEDIR/hoptimator-cli/build/install/hoptimator-cli/bin/hoptimator-cli \
   sqlline.SqlLine \
   -ac sqlline.HoptimatorAppConfig \
   --verbose \
-  -u "jdbc:hoptimator://fun=mysql;hints=path=ads_offline,job.properties.account=foo" -n "" -p "" -nn "Hoptimator" $@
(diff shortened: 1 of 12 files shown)


hoptimator
```diff
@@ -8,5 +8,5 @@ $BASEDIR/hoptimator-cli/build/install/hoptimator-cli/bin/hoptimator-cli \
   sqlline.SqlLine \
   -ac sqlline.HoptimatorAppConfig \
   --verbose \
-  -u "jdbc:hoptimator://fun=mysql;hints=offline.table.name=ads_offline,job.properties.account=foo" -n "" -p "" -nn "Hoptimator" $@
+  -u "jdbc:hoptimator://fun=mysql;hints=path=ads_offline,job.properties.account=foo" -n "" -p "" -nn "Hoptimator" $@
```
ryannedolan: Shouldn't need this now right?

ryannedolan (commented):
Nice! Couple comments.

hoptimator
srnand: correct, let me fix.

ryannedolan (approved):
nice!

PR #225, opened 2026-05-18
Support WITH options in FIRE TRIGGER

srnand:
## Summary
Extend FIRE TRIGGER with an optional WITH (k v, ...) options clause whose pairs are merged into the trigger's spec.jobProperties before the fire timestamp is bumped. The deployer rejects FIRE on an in-flight execution (status.timestamp > status.watermark) so callers see explicit failure rather than racing two triggered Jobs.

Option keys with the job.properties. prefix have it stripped on the way into spec.jobProperties, mirroring CREATE TRIGGER. Unprefixed keys are stored verbatim. The new Trigger.FIRE_OPTION marker carries the fire intent through DeploymentService to K8sTriggerDeployer.

Why: enables FIRE TRIGGER to inject per-fire parameters (e.g. backfill window bounds) into the trigger-job pod without baking domain-specific options into the SQL grammar.

### DIFF
--- a/hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorDdlExecutor.java
+++ b/hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorDdlExecutor.java
@@ -31,6 +31,7 @@ import com.linkedin.hoptimator.jdbc.ddl.SqlCreateMaterializedView;
 import com.linkedin.hoptimator.jdbc.ddl.SqlCreateTable;
 import com.linkedin.hoptimator.jdbc.ddl.SqlCreateTrigger;
 import com.linkedin.hoptimator.jdbc.ddl.SqlDropTrigger;
+import com.linkedin.hoptimator.jdbc.ddl.SqlFireTrigger;
 import com.linkedin.hoptimator.jdbc.ddl.SqlPauseTrigger;
 import com.linkedin.hoptimator.jdbc.ddl.SqlResumeTrigger;
 import com.linkedin.hoptimator.util.DeploymentService;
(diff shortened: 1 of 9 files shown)


hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorDdlExecutor.java
```diff
@@ -311,6 +312,41 @@ public void execute(SqlResumeTrigger resume, CalcitePrepare.Context context) {
...
+      throw new DdlException(fire, e.getMessage(), e);
+    }
+
+    if (fire.name.names.size() > 1) {
+      throw new DdlException(fire, "Triggers cannot belong to a schema or database.");
+    }
+    String name = fire.name.names.get(0);
+
+    Map<String, String> options = HoptimatorDdlUtils.options(fire.options);
+    options.put(Trigger.FIRE_OPTION, "true");
```
ryannedolan, line 333: So the difference between `create trigger` and `fire trigger` is essentially this `fire: true`? Wonder what happens if you `create trigger ... with (fire 'true')`?

ryannedolan (approved):
neat!

hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorDdlExecutor.java
srnand: As of now, `WITH (fire 'true')` on `create trigger` doesnt set `Trigger.FIRE_OPTION`. We could wire it up easily but i think it will get little more complex, 
 1. If create trigger succeeds but the fire step fails, do we roll back the CR?
 2. `CREATE OR REPLACE ... WITH (fire 'true')` needs to decide if we fire only on replace, only on create, or both?

So, i feel we should keep the firing a trigger as a separate grammar for itself.

PR #228, opened 2026-05-21
Use RouterBasedStoreMetadataFetcher for Venice store discovery

srnand:
## Summary
- Switched Venice JDBC store discovery from `ControllerClient` based, cluster-by-cluster listing to `RouterBasedStoreMetadataFetcher` (single router `/stores` call). 
- This PR takes the approach over the OSS HTTP/HTTPS route, which is the only transport available in non-D2 deployments.
- Bumps `venice-common` and `venice-thin-client` to `0.4.858`, which is the first version exposing `RouterBasedStoreMetadataFetcher`.

## Tests
- Added UTs.


ryannedolan (approved):
Nice!

PR #230, opened 2026-05-21
Add apply mode for declarative, idempotent CREATE DDL

ryannedolan:
## Summary
- Adds a `mode` JDBC connection property: `create` (default, strict) or `apply` (declarative, idempotent).
- In apply mode, plain `CREATE` converges the resource to the declared definition rather than failing on conflict — so checked-in `.sql` files become K8s-style reconcilable manifests for CI/CD.
- Routes all five CREATE forms (VIEW, MATERIALIZED VIEW, TABLE, TRIGGER, DATABASE) through a single `HoptimatorDdlUtils.effectiveMode(orReplace, conn)` helper that resolves the connection mode and the statement's `OR REPLACE` flag into a `DdlMode`.
- Existence-checks in `processCreateMaterializedView` / `processCreateTable` now key off the resolved `DdlMode`. SPECIFY (dry-run) preserves its original syntax-driven preview so `!specify` still reflects what a real run would do.
- Data is never destroyed by any CREATE form. `DROP` stays imperative and metadata-only; prune-by-absence is explicitly out of scope.
- Destructive-metadata-change detection (FORCE semantics) is deferred to a follow-up; in apply mode today, `CREATE` and `CREATE OR REPLACE` are equivalent because the deployer can't yet distinguish compatible vs. incompatible diffs.

Usage:

```
jdbc:hoptimator://...;mode=apply
```

| Mode               | `CREATE`                              | `CREATE OR REPLACE`  |
|--------------------|---------------------------------------|----------------------|
| `create` (default) | Fail if the resource exists.          | Update if it exists. |
| `apply`            | Converge. Idempotent.                 | Same as `CREATE`.    |

## Testing Done
- [x] Unit tests added — 7 new JUnit5 tests for `effectiveMode` / `isApplyMode` covering default, explicit `create`, `apply`, case insensitivity, unknown values, and null properties.
- [x] One pre-existing unit test that bypassed the executor was updated to pass `DdlMode.UPDATE` directly when supplying `OR REPLACE` SQL — the helper now treats `DdlMode` as authoritative.
- [x] New Quidem integration script `k8s-apply-mode.id` exercising idempotent re-CREATE for views and tables, the OR-REPLACE-equals-CREATE invariant under apply, and DROP still working. Wired into `TestSqlScripts` via `run("k8s-apply-mode.id", "mode=apply")`.
- [x] `./gradlew :hoptimator-jdbc:test` — 508 tests pass.
- [x] `./gradlew compileTestJava` — all modules compile clean.
- [x] `make integration-tests` — deferred to CI (needs the K8s cluster).

## Docs
`docs/user-guide/ddl-reference.md` gets a new "CREATE semantics" section describing the two modes, the JDBC property, the data-safety guarantee, and the out-of-scope list.


hoptimator-jdbc/src/main/java/com/linkedin/hoptimator/jdbc/HoptimatorDdlUtils.java
```diff
@@ -89,6 +89,53 @@ public final class HoptimatorDdlUtils {
...
+   * <ul>
+   *   <li>{@code create} (default) — strict create. {@code CREATE} fails if the resource
+   *       already exists; {@code CREATE OR REPLACE} updates.</li>
+   *   <li>{@code apply} — declarative, K8s-style reconciliation. Both {@code CREATE} and
+   *       {@code CREATE OR REPLACE} converge the resource to the declared definition, so
+   *       running the same script twice is a no-op. Idempotent by design.</li>
+   * </ul>
+   *
+   * <p>The property may also be set per statement via {@code SET mode = 'apply'} on a
+   * compatible session — see {@code docs/user-guide/ddl-reference.md}.
```
jogrogan: Is this true? Where does this get wired in?

jogrogan (commented):
Hmm, so this would effectively be a configuration on the jdbc end? We would silently set mode=apply and then it won't matter to users if they issue CREATE or CREATE OR REPLACE anymore as they'd have the same behavior?

ryannedolan:
> this would effectively be a configuration on the jdbc end

yes, that's the idea. This will enable us to build "as code" tooling which works similar to `kubectly apply`.

jogrogan (approved):
Nice!

PR #232, opened 2026-06-15
Fix invalid K8s label value error

jogrogan:
Fixes the following error we ran into:
```
a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?'
```

Issue is that the label appender was truncating labels blindly over 64 characters with no regards to the new ending character.

### DIFF
--- a/hoptimator-cli/src/main/java/sqlline/HoptimatorAppConfig.java
+++ b/hoptimator-cli/src/main/java/sqlline/HoptimatorAppConfig.java
@@ -23,7 +23,6 @@ import org.apache.calcite.util.Pair;
 import org.jline.reader.Completer;
 
 import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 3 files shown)


srnand (approved):

ryannedolan (commented):
For follow up -- probably should move this to K8sUtils: https://github.com/linkedin/Hoptimator/blob/4f2728c7ea520f3663f167682ddc7277d7b79f66/hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sUtils.java#L32

PR #235, opened 2026-06-17
Add dry-run via `deploy=false` connection property

ryannedolan:
## Summary

 - New `deploy` connection property: when `deploy=false`, each DDL statement is parsed, validated, and
  applied to the in-memory Calcite schema, but the underlying deployers are **not** invoked. Lets a session
   validate a multi-statement script end-to-end (later statements see earlier ones' in-memory effects)
  without external side-effects.
 - Orthogonal to `mode`: combine `deploy=false` with `mode=apply` to dry-run an apply-mode script.
 - Distinct from `SPECIFY` (`!specify`), which remains the strict zero-side-effect single-statement
  preview that still invokes `deployer.specify()` and unwinds the in-memory schema afterward. Dry-run
  preserves the in-memory mutation across statements and invokes no deployer method at all.

## Details

 - `HoptimatorDdlUtils.isDryRun(Connection)` reads the `deploy` property (defaults to `true`/live).
 - `DdlMode.CREATE` and `DdlMode.UPDATE` branches gate the `DeploymentService.create/update` call on
  `isDryRun`; `SPECIFY` is unchanged.  - Imperative non-CREATE DDL paths (`DROP {TABLE, VIEW, MATERIALIZED VIEW, TRIGGER}`, `FIRE/PAUSE/RESUME
  TRIGGER`) use inline `if (!isDryRun(connection))` guards. These don't fit the `DdlMode` enum, which is
  about CREATE-statement semantics (strict vs. apply vs. preview).
 - `CREATE VIEW` and `CREATE TRIGGER` previously bypassed `DdlMode.executeDeployers` and called
  `DeploymentService.create/update` directly; routed through `mode.executeDeployers` to inherit the gating.

## Testing Done

```
0: Hoptimator> create view bar as select * from foo;
INFO Validating statement: CREATE VIEW `BAR` AS
SELECT *
FROM `FOO`
INFO Validated sql statement. The view is named BAR and has path [DEFAULT, BAR]
INFO Validating deployable resources for view BAR
INFO Validated view BAR
INFO Dry-run (deploy=false): skipping UPDATE of view BAR
INFO Added view BAR to schema DEFAULT
INFO CREATE VIEW BAR completed
No rows affected (0.212 seconds)
```

Notice the "dry run ... skipping" message in the logs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,9 +81,9 @@ deploy-kafka: deploy deploy-flink
 	kubectl wait --for=condition=Established=True crds/kafkas.kafka.strimzi.io
 	kubectl apply -f ./deploy/samples/kafkadb.yaml
 	kubectl apply -f ./deploy/dev/kafka.yaml
-	kubectl wait kafka.kafka.strimzi.io/one --for=condition=Ready --timeout=10m -n kafka
-	kubectl wait kafkatopic.kafka.strimzi.io/kafka-database-existing-topic-1 --for=condition=Ready --timeout=10m
-	kubectl wait kafkatopic.kafka.strimzi.io/kafka-database-existing-topic-2 --for=condition=Ready --timeout=10m
+	kubectl wait kafka.kafka.strimzi.io/one --for=condition=Ready --timeout=15m -n kafka
+	kubectl wait kafkatopic.kafka.strimzi.io/kafka-database-existing-topic-1 --for=condition=Ready --timeout=15m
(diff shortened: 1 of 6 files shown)


jogrogan (approved):
LGTM!

ryannedolan:
Copilot fixed the build but clobbered the PR in the process. Sorry for the churn.

PR #237, opened 2026-07-01
Add data-availability triggering for TableTriggers

ryannedolan:
# Summary

 - `InputWatermarkSource` capability for `Databases`.
 - `FIRE TRIGGER ... FROM ... TO`
 - _ad hoc_ backfill support
 - Automatic late data repair via automatic backfills
 - Demo Kafka event source

# Details

Introduces a uniform way for `TableTriggers` to fire on arbitrary data change events. This is accomplished via a new interface `InputWatermarkSource`, which enables adapters to provide change events for specific storage systems. The Kafka adapter is extended as a basic example.

Supports on-demand backfills and automatic repair of late/out-of-order data. On-demand backfills can be triggered with `FIRE TRIGGER...FROM...TO`. Automatic backfills are scheduled by the reconciler when late data is detected (via `InputWatermarkSource`). This means that `TableTriggers` will fire under 3 situations:

 1. *New* data arrives. This existed prior to this PR, and represents the normal operation of `TableTriggers`.
 2. _Ad hoc_ backfills via `FIRE TRIGGER...FROM...TO`. These are just one-off `Jobs`.
 3. *Late* data arrives -- data that represents changes _before_ the watermark.

All 3 flows are managed by a new source-agnostic reconciler. (Previously, this was done with bespoke reconcilers for specific sources in our internal deployments.)

# Testing Done

Creating a trigger:

```
0: Hoptimator> create trigger foo on ads.page_views as 'retl-job-template' with ('job.properties.online.table.name' 'foo');
INFO Validating statement: CREATE TRIGGER `FOO` ON `ADS`.`PAGE_VIEWS` AS 'retl-job-template' WITH ('job.properties.online.table.name', 'foo')
INFO Validating trigger FOO with deployers
INFO Validated trigger FOO
INFO Creating trigger FOO
INFO Successfully snapshot K8s obj: TableTrigger:foo
INFO Created K8s obj: TableTrigger:foo
INFO Deployed trigger FOO
INFO CREATE TRIGGER FOO completed
No rows affected (0.1 seconds)
```

_Ad hoc_ firing a trigger:

```
0: Hoptimator> fire trigger foo;
INFO Validating statement: FIRE TRIGGER `FOO`
INFO Firing trigger FOO
INFO Updated K8s obj status: TableTrigger:foo
INFO FIRE TRIGGER FOO completed
No rows affected (0.082 seconds)
```

```
$ k get jobs -w
NAME                     STATUS     COMPLETIONS   DURATION   AGE
foo-retl-job-template-job   Running    0/1                      0s
foo-retl-job-template-job   Running    0/1           0s         0s
foo-retl-job-template-job   Running    0/1           3s         3s
foo-retl-job-template-job   Complete   1/1           3s         3s
```

Backfill:

```
0: Hoptimator> fire trigger foo from 2 days ago to now;
INFO Validating statement: FIRE TRIGGER `FOO` FROM '-2d' TO 'now'
INFO Firing trigger FOO
INFO Capping backfill end for trigger FOO from 2026-07-15T03:32:26Z to the watermark 2026-07-15T03:27:39.141872Z (a backfill cannot cover data the cursor has not yet processed).
INFO Created K8s obj: Job:foo-retl-job-template-job-bf-fc39090c
INFO FIRE TRIGGER FOO completed
No rows affected (0.096 seconds)
```

```
$ k logs foo-retl-job-template-job-bf-f59ce552-gnhjf
foo-retl-job-template-trigger processing window [2026-07-13T04:21:08Z, 2026-07-15T04:19:36.962060Z] at Wed Jul 15 04:21:09 UTC 2026
```


hoptimator-kafka/src/main/java/com/linkedin/hoptimator/kafka/ClusterSchema.java
```diff
@@ -71,4 +84,45 @@ protected String getDescription() {
...
+          adminClient.describeTopics(Collections.singleton(topic)).allTopicNames().get().get(topic);
+      if (description == null) {
+        return Optional.empty();
+      }
+      Map<TopicPartition, OffsetSpec> query = new HashMap<>();
+      for (TopicPartitionInfo partition : description.partitions()) {
+        query.put(new TopicPartition(topic, partition.partition()), OffsetSpec.maxTimestamp());
+      }
+      long frontier = Long.MIN_VALUE;
+      for (ListOffsetsResultInfo info : adminClient.listOffsets(query).all().get().values()) {
```
srnand, line 130: Shouldn’t the completeness watermark be conservative and use the minimum of all partition's latest timestamps?

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
```diff
@@ -241,6 +449,97 @@ private ExecutionTime scheduledExecution(V1alpha1TableTrigger object) {
...
+
+  /**
+   * Consumes the source's change stream (in arrival order, via {@code status.lateWatermark}) and, on
+   * the first change whose data-time window lies behind the watermark, enqueues a one-off backfill
+   * over that window — clipped to the watermark so it never runs ahead of the cursor. Returns a
+   * requeue {@link Result} when it writes status (a repair enqueued or the cursor consumed), or null
+   * when there is nothing to do. On first sight (no {@code lateWatermark}) it initializes the cursor
+   * to now, so a freshly-created trigger reacts only to <em>future</em> late changes, not all history.
+   */
+  private Result maybeEnqueueLateRepair(V1alpha1TableTrigger object, V1alpha1TableTriggerStatus status)
```
srnand, line 339: Here the lateWatermark is advanced on a new backfill job but if that job fails, the clearBackfill clears backfillFrom/To but leaves the lateWatermark advanced. So seems like a late change whose backfill fails is silently lost, with no retry and nothing recorded on the CRD ?

srnand (commented):
niceee!

hoptimator-kafka/src/main/java/com/linkedin/hoptimator/kafka/ClusterSchema.java
ryannedolan: Hmm. I think Claude is trying to fire the trigger anytime any data arrives in a Kafka topic, but then why is it called "watermark" and not "frontier"?

ryannedolan: confirmed; should be named `frontier`. This gets confusing. I think `Input*Watermark*Source` confused us.

srnand: Gotcha but I am still worried that the frontier can cause missed records. 

If partition-0 sets the watermark to T0 (the highest timestamp) when the job finishes, and later partition-1 produces a record with timestamp T1 where T1 < T0, that record won’t be processed because the watermark has already passed it. Only a manual backfill would include it.

ryannedolan: Yeah. I think Kafka ends up being a very poor source for this sort of thing. We need a monotonically increasing T, but with Kafka we cannot make that guarantee across partitions. This PR would treat T1 as late-arriving data, which makes sense for batch workflows but not really for Kafka.

Lemme try to add a lagging lookback window.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
ryannedolan: Yeah, I think Claude is trying too hard to re-create the Job controller (having seen how we have similar complex controllers internally). Lemme see if we can just use the Job controller.

hoptimator-kafka/src/main/java/com/linkedin/hoptimator/kafka/ClusterSchema.java
ryannedolan: Added a lagging window, which I think is as good as we can get here. Basically: a Kafka _partition_ that is idle is ignored until new data arrives, rather than hold back the entire stream. Switched to use `min`, so we conservatively track all (non-idle) partitions.

hoptimator-operator/src/main/java/com/linkedin/hoptimator/operator/trigger/TableTriggerReconciler.java
ryannedolan: Cleaned a lot of this up. Now _ad hoc_ jobs are created directly from the `Deployer` rather than jumping through the CRD and controller. Means we don't need to track as much state, and we don't need to re-create a `Job` controller.

ryannedolan: wrt lost backfills, now each backfill is a `Job` which will eventually complete (or `Fail`). The `Job` controller (K8s built-in) will keep trying for us, and we can configure a max retries etc if we wish.

srnand (approved):

hoptimator-kafka/src/main/java/com/linkedin/hoptimator/kafka/ClusterSchema.java
srnand: yeaa there could still be problems with the lagging window. if one of the partition is a hot partition, all the others would be marked as idle. But agreed, kafka is not a great source for this use case.

ryannedolan: I thought the same, but it's a bit better than that. If we have one hot partition and N-1 idle partitions, that's actually totally fine, because it means *no records* have been sent on N-1 partitions. As soon as even one record is sent to an idle partition, the partition will be considered part of the non-idle set and affect the frontier (assuming that the timestamp is close to _now_, which ofc is usually the case for new records).

The key here is that we aren't actually *consuming* these records. We're just observing the greatest timestamp in each partition as reported by the broker. So we don't need to worry about "lagging" partitions in the usual Kafka sense (i.e. a consumer cannot keep up).

The big gap I see is that we totally ignore records with old timestamps. But it would be probably worse if we tried to trigger a backfill every time old records came through. That probably isn't what Kafka consumers would expect.

srnand: ahh yess, that makes sense 💡

PR #240, opened 2026-07-27
SPI foundation: replace java.sql.Connection with DeploymentContext

jogrogan:
# Decouple Proteus DDL
Hoptimator's validation and deployment logic is currently reachable only by issuing SQL DDL through a Calcite/JDBC connection. That couples every caller to Calcite and forces schema definitions through a lossy  Avro → RelDataType → Avro  round-trip. Hoptimator via a CLI or over REST/gRPC may want to create/validate tables by handing us a table name + a schema directly — not by constructing DDL.

This stacked series decouples the validators and deployers from Calcite so the same validation and deployment engine can be driven either by SQL or by a direct, connection-free call — while keeping validators/deployers as a clean SPI. The payoff: an API call to "create table from a schema" path with identical guarantees to DDL (including dry-run), lossless Avro fidelity, and no Calcite dependency for callers that don't want one. It also keeps the door open for follow-ups (e.g. materializations, source→sink field mappings) without re-plumbing.

Stacked series (review/merge in order):
1.  decouple-pr1-spi  — SPI foundation:  java.sql.Connection  →  DeploymentContext  (pure decoupling, no behavior change)
2.  decouple-pr2-api  — SQL-free  TableService  + lossless Avro preservation
3.  decouple-pr3-hardening  — typed transient failures + required  Deployer.exists() 
(This is PR 1 of 3.)

##  PR 1/3 — SPI foundation: replace `java.sql.Connection` with `DeploymentContext`

Today the deploy/validate SPI is threaded with `java.sql.Connection`, which hard-wires the
control plane to Calcite/JDBC. That makes it impossible to drive table validation and
creation any way other than through a JDBC connection running DDL. This PR introduces a
neutral seam — `DeploymentContext` — and moves the entire SPI and every deployer off
`java.sql.Connection`, **without changing any behavior**. It is pure decoupling: the
SQL/DDL path works exactly as before, just over the new context instead of a connection.

This is the groundwork for a SQL-free, connection-free table API (PR2), which in turn
lets callers (e.g. a gRPC service) request the same validation + deployment from a table
path and a schema rather than by issuing DDL.

## What changes

- **New SPI type `DeploymentContext`** (`hoptimator-api`) — the connection-free handle the
  deploy/validate machinery now depends on, replacing `java.sql.Connection` across
  `Deployer`, `Validator`, `Connector`, and the `*Provider` interfaces.
- **`CalciteDeploymentContext`** — the SQL-path implementation that adapts the existing
  Calcite/JDBC world onto `DeploymentContext`, so the DDL path is unchanged.
- **`DualLogger`** extracted from `HoptimatorConnection` — log-hook fan-out is no longer
  tied to the connection object.
- **`SimpleDeploymentContext`** — a minimal context for connection-free control-plane
  callers.
- **`databaseProperties(...)`** now takes optional `catalog`/`schema`, so config resolution
  no longer assumes a fully-qualified JDBC identifier.
- **All deployers migrated** off `java.sql.Connection` — K8s (and its YAML/job/source/MV
  deployers), Kafka, MySQL, Venice, and the logical-table deployer — plus the
  `Config`/`Connector`/`Validator` providers.
- `HoptimatorDdlUtils` refactored to drive deployment through the context rather than the
  connection.

## What this PR does **not** do

- No direct/SQL-free table API yet (that's PR2).
- No Avro schema preservation (PR2).
- No behavior changes to validation, deployment, or DDL semantics.

## Testing

- Unit suite green (deployer/provider/service tests updated to the new context; new
  `DualLoggerTest`, `SimpleDeploymentContextTest`, `TestDeploymentContext`).
- Existing integration behavior unchanged — the SQL/DDL path is exercised as before.

## Review notes

- ~118 files, but the vast majority are mechanical `Connection → DeploymentContext`
  signature swaps in deployers/providers and their tests.
- The conceptual core is `DeploymentContext`, `CalciteDeploymentContext`, `DualLogger`, and
  the `HoptimatorDdlUtils` refactor; the rest follows from those.
- Docs updated: `docs/extending/validators.md`.

### DIFF
--- a/hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sDatabaseTable.java
+++ b/hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sDatabaseTable.java
@@ -1,5 +1,6 @@
 package com.linkedin.hoptimator.k8s;
 
+import com.linkedin.hoptimator.DeploymentContext;
 import com.linkedin.hoptimator.jdbc.HoptimatorConnection;
 import com.linkedin.hoptimator.k8s.models.V1alpha1Database;
 import com.linkedin.hoptimator.k8s.models.V1alpha1DatabaseList;
@@ -54,16 +55,17 @@ public class K8sDatabaseTable extends K8sTable<V1alpha1Database, V1alpha1Databas
   }
 
   public void addDatabases(SchemaPlus parentSchema, Connection connection) {
+    DeploymentContext context = ((HoptimatorConnection) connection).deploymentContext();
     for (Row row : rows()) {
       if (row.CATALOG != null) {
         Schema catalogSchema = HoptimatorJdbcCatalogSchema.create(row.NAME, row.CATALOG, row.SCHEMA, dataSource(row,
                 ((HoptimatorConnection) connection).connectionProperties()), parentSchema,
-            dialect(row), engines.forDatabase(row.NAME), connection);
+            dialect(row), engines.forDatabase(row.NAME), context);
         parentSchema.add(row.CATALOG.toUpperCase(Locale.ROOT), catalogSchema);
       } else {
         Schema schema = HoptimatorJdbcSchema.create(row.NAME, row.CATALOG, row.SCHEMA, dataSource(row,
                 ((HoptimatorConnection) connection).connectionProperties()), parentSchema,
-            dialect(row), engines.forDatabase(row.NAME), connection);
+            dialect(row), engines.forDatabase(row.NAME), context);
         parentSchema.add(schemaName(row), schema);
       }
     }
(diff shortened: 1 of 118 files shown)


hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sDatabaseTable.java
```diff
@@ -101,30 +108,39 @@ static String schemaName(Row row) {
...
-        String value = connectionProperties.getProperty(key);
-        joiner.add(key + "=" + value);
+      }
+    }
+    return JdbcSchema.dataSource(joinedUrl(row, connectionProperties), row.DRIVER, user, pass);
+  }
+
+  /**
+   * Builds the effective JDBC URL for a Database: its CRD {@code url} with the connection-level
+   * properties (except {@code user}/{@code password}) and the CRD name appended as
```
ryannedolan, line 123: Claude loves to conflate CRD with CR (everyone does).

ryannedolan (approved):
love it

hoptimator-k8s/src/main/java/com/linkedin/hoptimator/k8s/K8sDatabaseTable.java
jogrogan: Yea that's fair, did a quick check and it seems to be conflating in a bunch of spots, many already checked in. I'll try to clean it up separate from this.

PR #248, opened 2026-08-11
Handle dotted table names in quoted identifiers

jogrogan:
## Problem

A quoted identifier segment can legitimately contain a dot — e.g. the Kafka topic `"KAFKA"."my.event"`. Calcite parses such statements correctly (a quoted segment stays whole), but several Hoptimator paths took a **bare identifier string** and naively `split("\\.")`, shredding `my.event` into `my` + `event`. That broke:

- `!describe` / `!graph` / `!resolve` CLI commands
- `GraphService.resolve` (graph target resolution)
- `PipelineGraphBuilder` custom-resource name lookups

## Fix

New `IdentifierUtils.parseIdentifier(String)` that splits by quoting:

- **Unquoted input** → split directly on `.`. This is exact (an unquoted segment can't contain a dot) and also covers the unquoted hyphenated names the CLI commands accept but the SQL grammar rejects, e.g. `LOGICAL.testevent-graph` (which the parser would read as subtraction).
- **Quoted input** → uses the SQL parser so a dot *inside* a quoted segment (`"my.table"`) is preserved. Malformed/mixed quoted forms (e.g. `"KAFKA".my-topic`) now throw rather than silently mis-splitting.

Wired into `GraphService.resolve`, the `!resolve`/`!describe` CLI + quidem paths, and `PipelineGraphBuilder.forView`/`forLogicalTable`.

**Naming is intentionally unchanged.** K8s object names here are DNS-1123 subdomains and already permit dots, and resources with dotted names exist in production. `canonicalizeName` is left as-is so derived names stay backwards compatible (create-time and lookup-time both yield `venice-my.event`); rewriting dots would make the operator create duplicates instead of updating existing resources.

## Tests

- `IdentifierUtilsTest` — quoted dotted segments, escaped quotes, unquoted/hyphenated, and malformed-quoted-throws cases.
- `K8sUtilsTest` — dot-preservation cases documenting the backwards-compatible naming behavior.
- `GraphServiceTest` — resolves `"VENICE"."my.table"` as a single table identity.
- `hoptimator-kafka` integration script `kafka-ddl-dotted.id` — creates, `!describe`s, and drops a topic whose name contains a dot (verified against a live cluster).

Unit suites, spotbugs, checkstyle, and the affected integration tests all pass.

### DIFF
--- a/hoptimator-k8s/src/test/java/com/linkedin/hoptimator/k8s/K8sUtilsTest.java
+++ b/hoptimator-k8s/src/test/java/com/linkedin/hoptimator/k8s/K8sUtilsTest.java
@@ -42,6 +42,20 @@ class K8sUtilsTest {
   }
 
   @Test
+  void canonicalizeNamePreservesDots() {
+    // Dots are intentionally NOT rewritten: K8s object names here are DNS-1123 subdomains, which
+    // permit dots, and existing deployed resources already carry dotted names. Rewriting dots
+    // would change the derived name and make the operator create duplicates instead of updating
+    // the pre-existing resource. Keeping dots keeps create-time and lookup-time names in agreement.
+    assertEquals("my.event", K8sUtils.canonicalizeName("my.event"));
+  }
+
+  @Test
+  void canonicalizeNameWithDottedTableAndDatabasePreservesDot() {
+    assertEquals("kafka-database-my.event", K8sUtils.canonicalizeName("kafka-database", "my.event"));
+  }
+
+  @Test
   void canonicalizeNameWithDatabaseAndTable() {
     assertEquals("mydb-mytable", K8sUtils.canonicalizeName("myDB", "myTable"));
   }
--- a/hoptimator-cli/src/main/java/sqlline/HoptimatorAppConfig.java
+++ b/hoptimator-cli/src/main/java/sqlline/HoptimatorAppConfig.java
@@ -11,6 +11,7 @@ import com.linkedin.hoptimator.jdbc.HoptimatorDriver;
 import com.linkedin.hoptimator.jdbc.ResolvedTable;
 import com.linkedin.hoptimator.jdbc.ddl.SqlCreateMaterializedView;
 import com.linkedin.hoptimator.util.DeploymentService;
+import com.linkedin.hoptimator.util.IdentifierUtils;
 import com.linkedin.hoptimator.util.planner.PipelineRel;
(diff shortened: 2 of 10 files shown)


hoptimator-k8s/src/test/java/com/linkedin/hoptimator/k8s/K8sUtilsTest.java
```diff
@@ -41,6 +41,20 @@ void canonicalizeNameReplacesUnderscoresAndDollarSigns() {
     assertEquals("my-table", K8sUtils.canonicalizeName("my_$table"));
   }
 
+  @Test
+  void canonicalizeNamePreservesDots() {
+    // Dots are intentionally NOT rewritten: K8s object names here are DNS-1123 subdomains, which
```
ryannedolan, line 46: That might be technically true, but I have never seen periods in CR names, and some linters will complain about them. I think it's probably safer if we replace periods with hyphens.

hoptimator-k8s/src/test/java/com/linkedin/hoptimator/k8s/K8sUtilsTest.java
```diff
@@ -41,6 +41,20 @@ void canonicalizeNameReplacesUnderscoresAndDollarSigns() {
...
+    // permit dots, and existing deployed resources already carry dotted names. Rewriting dots
+    // would change the derived name and make the operator create duplicates instead of updating
+    // the pre-existing resource. Keeping dots keeps create-time and lookup-time names in agreement.
+    assertEquals("my.event", K8sUtils.canonicalizeName("my.event"));
+  }
+
+  @Test
+  void canonicalizeNameWithDottedTableAndDatabasePreservesDot() {
+    assertEquals("kafka-database-my.event", K8sUtils.canonicalizeName("kafka-database", "my.event"));
```
ryannedolan, line 55: typical order of operations makes this weird to read

hoptimator-k8s/src/test/java/com/linkedin/hoptimator/k8s/K8sUtilsTest.java
jogrogan: I fact checked and unfortunately we have some CRs in prod that use dots in the name. Changing this behavior would effectively orphan them, they’d no longer be able to be updated or dropped and the same view reissued would lead to a duplicate job.

ryannedolan (approved):

PR #250, opened 2026-09-04
Validate source/sink types and inject assignment casts on write

jogrogan:
## Summary

When a materialized view or `INSERT` writes into a sink with a declared schema, Hoptimator previously deferred all query-vs-sink type checking to the Flink job, so a mismatch — e.g. a raw Kafka primitive key exposed as `STRING` projected onto a typed `BIGINT` key column — only failed at job submission with a cryptic Flink error. This adds plan-time reconciliation in `PipelineRel`: Hoptimator now compares each projected column to its sink column and either injects an assignment `CAST` or raises a clear error up front.

## Behavior

A new `castMode` hint (fail-closed, defaults to `strict`) controls how aggressively casts are injected. It maps directly onto Calcite's `SqlTypeUtil.canCastFrom`:

| Mode | Injects a cast for | Notes |
|------|--------------------|-------|
| `strict` (default) | implicitly-assignable (lossless) conversions, e.g. widening `INTEGER -> BIGINT` | rejects `STRING -> BIGINT` and lossy narrowing `BIGINT -> INTEGER` |
| `assign` | `strict` + a character-to-scalar carve-out (covers the raw key `STRING -> BIGINT`) | |
| `explicit` | any explicitly-castable scalar pair | the deliberate opt-in, incl. lossy narrowing |

Rules that hold at every mode:
- Same-family differences (VARCHAR length, DECIMAL precision, nullability) are assigned as-is and enforced by the engine — only genuine cross-family mismatches are cast.
- A nullable source into a `NOT NULL` sink column **that also needs a cast** is rejected (a cast can't add a null guard); matching types pass through regardless of nullability.
- Complex `ROW`/`ARRAY`/`MAP`/`MULTISET` shape mismatches are always rejected.
- An explicit user `CAST`/`SAFE_CAST` is respected and never double-wrapped; if its result type isn't assignable to the sink, we fail early with a targeted message.

## Implementation

- `TypeCoercion` (new): `CastMode` + `decide(source, target, mode)` classifier over Calcite's assignment/cast rules.
- `PipelineRel.Implementor.sql()`: `resolveCasts()` validates the query against the sink and returns per-column cast targets, throwing `SQLNonTransientException` early on incompatibility.
- `ScriptImplementor`: threads per-column cast targets into the sink-aligned forced projection; all prior `insert()` overloads are preserved (backward compatible).

Note on scope: this governs the `CREATE MATERIALIZED VIEW` path (where Hoptimator resolves the sink separately, so Calcite does not coerce — the original failure). Plain `INSERT INTO <existing table>` is already coerced by Calcite's validator, so `castMode` is effectively a no-op there.

## Docs

- `docs/user-guide/hints.md` — the `castMode` planner hint.
- `docs/user-guide/ddl-reference.md` — "Type checking and casts on write".

## Testing

- **Unit** — `TypeCoercionTest` (single test file per source file): exact scalar matrix (numeric widening/narrowing, character coercions, temporal, binary, boolean), complex-type matrix, and cartesian property tests (mode monotonicity strict subset-of assign subset-of explicit, reflexivity, complex-vs-scalar, nullability). ~1400 cases.
- **Integration** — Quidem `.id` fixtures verified end-to-end against a live environment:
  - `hoptimator-venice`: `venice-ddl-cast.id` (assign injects `CAST` into a Venice keyed sink).
  - `hoptimator-k8s`: strict reject, user-CAST respected (no double-wrap), user-CAST-not-assignable error, matching-type/nullable passthrough, assign auto-cast + `VARBINARY`/nullable rejections, explicit-only casts.

No golden-fixture changes for existing pipelines; matching-type passthroughs are unaffected.


hoptimator-k8s/src/test/resources/k8s-cast-assign.id
```diff
@@ -0,0 +1,51 @@
...
+    - CREATE TABLE IF NOT EXISTS `ADS`.`cast_asgn` (`member_id` BIGINT, `name` VARCHAR, `vb` VARBINARY) WITH ('connector'='blackhole')
+    - INSERT INTO `ADS`.`cast_asgn` (`member_id`, `name`) SELECT CAST(`MEMBER_URN` AS BIGINT) AS `member_id`, `FIRST_NAME` AS `name` FROM `PROFILE`.`MEMBERS`
+    jarURI: file:///opt/hoptimator-flink-runner.jar
+    parallelism: 1
+    upgradeMode: stateless
+    state: running
+!specify cast_asgn
+
+create or replace materialized view ads."cast_asgn$to-binary" ("vb") as select member_urn AS "vb" from profile.members;
+Incompatible types for sink column 'vb': query produces VARCHAR but sink expects VARBINARY (no safe conversion is available under castMode=assign).
```
ryannedolan, line 36: neat!

ryannedolan (approved):

PR #252, opened 2026-09-16
Evaluate template conditional guards before expanding other variables

jogrogan:
## Problem

A `JobTemplate`/`TableTemplate` guarded with `{{var==value}}` (or `{{var!=value}}`) is meant to render only when the condition holds and be skipped entirely otherwise. But `SimpleTemplate.render()` expanded `{{...}}` placeholders in **text order**, so a guard placed *after* another variable did not protect it — the earlier variable was resolved first.

This breaks the "two templates with mirroring guards" pattern. With a Flink SQL job template and a Flink Beam job template both installed (guarded on `li.flink.app.type==SQL` and `==BEAM`), deploying a **Beam** job should skip the SQL template. But because `{{flinksql}}` appears before its `{{li.flink.app.type==SQL}}` guard, the SQL body was generated first and could **throw** (e.g. failing type validation on a cast the Beam path doesn't care about), breaking the deployment instead of cleanly skipping the template.

## Fix

Parse the template **once** into an ordered list of literal/placeholder tokens, then render in two passes over that list:

1. Evaluate every conditional guard first, regardless of position — if any fails, skip the whole template (`return null`) before expanding anything else.
2. Expand the remaining `{{var}}` / `{{var:default}}` values.

Guards can now sit anywhere in the template and reliably protect expensive or fail-prone variables.

The token model also lets rendering append directly to a `StringBuilder`, dropping the `Matcher.appendReplacement`/`quoteReplacement` escaping.

## Validation

- **Unit tests** (`TemplateTest`, +11): guard-before/after a throwing variable, multi-guard fail-fast, plus refactor-safety (`$`/`\` literal rendering, adjacent placeholders, empty template, guard-only template). All green.
- **Differential check** (throwaway): embedded the original single-pass algorithm and diffed it against the new one across the real `flink-template.yaml` / `flink-beam-template.yaml` shapes (multiline bodies, `$` in SQL, guards passing/failing) and edge inputs — **byte-for-byte identical**. The only intended behavior change is the guard ordering.
- `:hoptimator-util`, `:hoptimator-jdbc`, `:hoptimator-k8s` unit suites green.

## Docs

`docs/kubernetes/templates.md` — the "guard can sit anywhere" note now states the real guarantee: guards are evaluated before any other variable, so a guard reliably protects a variable that appears earlier in the text.


ryannedolan (approved):
interesting!

